PR #4, opened 2022-04-29
Add --trimpath to go tests

matthewmueller:
This is a follow-up from #2. The tests should have caught this issue, so I've updated the tests to catch this next time.


matthewmueller:
Closing since this is stale at this point.

PR #24, opened 2022-05-13
Fix for install URL to use Github link

fuziontech:
Current DNS is not resolving for the install link. Also having the url point to the file on Github is a bit more confidence inspiring. You can easily go look to see what is in the file without downloading it.

PS this project looks amazing! Excited to build on this.


matthewmueller:
Oof. I saw this after I already made the fix. Thanks for your contribution and kind words. Hit me up at hi@livebud.com if you have any questions after giving it a go!

PR #31, opened 2022-05-14
fix: move the prompt out of the routines

barelyhuman:
## Current Behavior 
Both npm output and the prompt for the app modules is overlapped by each other

## Changes by the PR 
Use the original context while prioritizing the prompt and then running the other stuff in the derived context and routines.

### DIFF
--- a/internal/command/create/create.go
+++ b/internal/command/create/create.go
@@ -35,7 +35,10 @@ func (c *Command) Run(ctx context.Context) error {
 		return err
 	}
 
-	c.generateGoMod(ctx, tmpDir)
+	err = c.generateGoMod(ctx, tmpDir)
+	if err != nil {
+		return err
+	}
 
 	eg, ctx2 := errgroup.WithContext(ctx)
 	eg.Go(func() error { return c.generateGitIgnore(ctx2, tmpDir) })
--- a/internal/command/create/create.go
+++ b/internal/command/create/create.go
@@ -34,10 +34,12 @@ func (c *Command) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	c.generateGoMod(ctx, tmpDir)
+
 	eg, ctx2 := errgroup.WithContext(ctx)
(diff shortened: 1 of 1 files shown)


matthewmueller:
Oh right, I remember seeing this once. This happens when you're working outside of `$GOPATH`. 

I'm wondering if the better solution would be to remove the NPM progress bar. That way it just installs in the background and doesn't clobber the go module question. There might be a flag you can pass to NPM, otherwise we can perhaps discard stdout/stderr.

What do you think?

barelyhuman:
I'm more towards keeping prompts out of the routines hence the above solution 
but yes, if you feel like we can have an ENV/flag to enable or disable npm logs, that works too

Though, the prompt being outside of routines while keeping the context shouldn't slow down the user since the other 2 tasks are still parallel in this case

internal/command/create/create.go
```diff
@@ -34,10 +34,12 @@ func (c *Command) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	c.generateGoMod(ctx, tmpDir)
```
matthewmueller, line 38: Let's check for the error and return the error if there's a failure.

matthewmueller (changes requested):

matthewmueller:
Ah I missed your earlier point. You're right, it doesn't make a ton of sense to have a prompt within a goroutine. Just reviewed, one fix and it's ready to merge. Thanks!

internal/command/create/create.go
barelyhuman: Updated

matthewmueller (approved):

matthewmueller:
Thanks @barelyhuman!

matthewmueller:
FYI. This is available in [v0.1.1](https://github.com/livebud/bud/releases/tag/v0.1.1). To upgrade, run:

```sh
curl -sf curl https://raw.githubusercontent.com/livebud/bud/main/install.sh | sh
```

PR #34, opened 2022-05-14
fix: installing binary with write permissions

barelyhuman:
## Current Behaviour
If the binary folder is not owned by the user then it fails to install it 

## After Merge 
The write permissions of the folder are checked before doing any installing
and elevated permissions are requested for it

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -47,10 +47,10 @@ execute() {
   (cd "${tmp_dir}" && untar "${TARBALL}")
 
   if [ -w "$BINDIR" ]; then
-  log_info "Installing $out to $prefix"
     install -d "${BINDIR}"
     install "${tmp_dir}/${NAME}/${BINARY}" "${BINDIR}/"
   else
+    echo "Elevated permissions required to install the binary"
     sudo install -d "${BINDIR}"
     sudo install "${tmp_dir}/${NAME}/${BINARY}" "${BINDIR}/"
   fi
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -45,10 +45,19 @@ execute() {
   hash_sha256_verify "${tmp_dir}/${TARBALL}" "${tmp_dir}/${CHECKSUM}"
 
   (cd "${tmp_dir}" && untar "${TARBALL}")
-  install -d "${BINDIR}"
-  install "${tmp_dir}/${NAME}/${BINARY}" "${BINDIR}/"
+
+  if [ -w "$BINDIR" ]; then
```
matthewmueller, line 49: Do you mind adding a comment above with what the conditional does? Otherwise everyone will need to google what `-w` does to understand this 😅

matthewmueller (changes requested):
Let's add a comment then this looks good to go. Thanks @barelyhuman!

install.sh
barelyhuman: Done 👍🏼

matthewmueller:
Thanks @barelyhuman!

matthewmueller:
FYI. This is available in [v0.1.0](https://github.com/livebud/bud/releases/tag/v0.1.1). To upgrade, run:

```sh
curl -sf curl https://raw.githubusercontent.com/livebud/bud/main/install.sh | sh
```

PR #59, opened 2022-05-16
readme: add background/philosophy section

thepudds:
Hi @matthewmueller, random gopher here. This project certainly seems to have a bunch of promise!

This PR attempts to capture some of the philosophy of the project that you described elsewhere, but is not currently in the README. 

The text is a slightly cut down version of what you posted to Reddit and Hacker News. (Empirically, that text seems to have been effective 😅 ).

My guess is that with more time, you would probably not use this text exactly (e.g., maybe your style is to not use the first person in a README), but I think having something like this sooner rather than later would be helpful for the project, and then it could be improved later. (And selfishly, I'd like to point some people to this repo, and it would be nice to have more context here).

I also included some pointers to other discussions at the bottom.

Happy for any feedback, or feel free to reject or tweak any or all of this!

Finally, I used the phrase "How did Bud come into existence?" for the new section header, but alternatives could be "Philosophy", "Background", "Why does this exist?", "Where did this come from?", or something else.


matthewmueller:
@thepudds – this is incredible. 

I love seeing you put this in your own words. It's much more concise than my rambling message earlier. I'm going to make a few tweaks, but otherwise I think this will be a very welcome addition to the readme. 

Thank you!!!

PR #64, opened 2022-05-17
Speed up tests by using existing GOMODCACHE

matthewmueller:
Slow test suites make you not want to write tests, but Bud is going to need a ton of tests.

One of the major reasons the test suite is slow is because we run some tests with a fresh `GOMODCACHE`. This was done to test plugin support, but these tests could easily take a minute to download modules.

This PR replaces using a custom `GOMODCACHE` in favor of real fixture modules: https://github.com/livebud/bud-test-plugin and https://github.com/livebud/bud-test-nested-plugin

### DIFF
--- a/package/watcher/watcher_test.go
+++ b/package/watcher/watcher_test.go
@@ -196,7 +196,7 @@ func TestWithScaffold(t *testing.T) {
 	is.True(paths["view/show.svelte"])
 	// While there should be no more events, testing this can be flaky in CI.
 	// Instead test that we don't have any events with unexpected paths.
-	// An extra event isn't the end of the world, it's just reload one more time.
+	// An extra event isn't the end of the world, it'll just reload one more time.
 	select {
 	case path := <-event:
 		rel, err := filepath.Rel(dir, path)


internal/generator/transform/transform_test.go
```diff
@@ -40,18 +39,7 @@ func TestMarkdownPlugin(t *testing.T) {
...
-				package markdown
-				import "github.com/livebud/bud/runtime/transform"
-				type Transform struct {}
-				func (t *Transform) MdToSvelte(file *transform.File) error {
-					file.Code = "<h1>hello</h1>"
-				}
-			`,
-		},
-	}
+	bud.Modules["github.com/livebud/bud-test-plugin"] = "v0.0.6"
```
matthewmueller, line 42: TODO make dynamic

matthewmueller:
The updated PR removes direct dependencies on the test plugins from Bud. Depending on these plugins directly was leading to the plugins seeping their way into tests in surprising ways. When testing, we use `replace` to run the tests with the latest runtime. This replace would then bring Bud's plugins into the tests.

There's something to be said about how many replaced modules shouldn't bring their own plugins. We'll need to see if this trips people up.

bdallhahmdsyry00 (commented):
#101

matthewmueller:
Not following @bdallhahmdsyry00

PR #66, opened 2022-05-18
Switch the tests to type-check

matthewmueller:
This PR is an experiment to switch the test suite to type-check instead of building and running Bud. This should hopefully speed up the test suite while continuing to catch the majority of errors.

We'd need to supplement this type-checking with end-to-end tests to catch logical errors, but those E2E tests can be more comprehensive.


matthewmueller:
I'm currently stuck on how to get the Go's type-checker to ignore errors from the standard library. I asked the question here: https://groups.google.com/g/golang-nuts/c/jcocEVMZIrM

If anyone has any idea, please chime in!

matthewmueller:
On second thought, running `go build` should catch these errors nearly as quickly with no extra work.

I think I'll just need to lean harder into binary caching and make it persistent in CI

PR #67, opened 2022-05-18
Support other Actions

### LINKED ISSUE #54: Support other Actions
Based on the docs:
image

When adding other actions currently it results in a 404.

theEyeD:
Fixes #54

Routes were already attempting be to registered but failing due to using the require initial uppercase letter of the func name. This change simply lowercases the route when registering.

### DIFF
--- a/package/router/radix/tree.go
+++ b/package/router/radix/tree.go
@@ -75,7 +75,7 @@ type tree struct {
 }
 
 func (t *tree) Insert(route string, handler http.Handler) error {
-	lexer := lex.New(route)
+	lexer := lex.New(strings.ToLower(route))
 	var tokens lex.Tokens
 	for {
 		token := lexer.Next()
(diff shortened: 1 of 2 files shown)


matthewmueller:
Thanks for giving this a shot Alex! 

I hope you don't mind, but I want to take it in a slightly different direction. The reason being that when we introduce custom routes, I don't really want developers to be able to add `router.Get("/Hi", hander)` and that gets mapped to `/hi`, I'd rather if fail faster.

I'd like the generators themselves to do more of the work. Your investigation helped me figure out what was going on so much faster, so thank you for that! I've opened the following PR: https://github.com/livebud/bud/pull/71/files.

theEyeD:
Sounds good, still trying to wrap my head around it all to know where these sorts of changes make sense 🙂

PR #68, opened 2022-05-19
204 improvements

### LINKED ISSUE #10: 204 No Content improvements
For example,

```go
func (c *Controller) Create() {
  // do something
}
```

These functions should return `204 No Content` if successful.

theEyeD:
Fixes #10 

* Return 204 on `nil` error responses
* Allow void view action by passing empty `map[string]interface{}{}` to view Handler

### DIFF
--- a/runtime/generator/controller/controller.gotext
+++ b/runtime/generator/controller/controller.gotext
@@ -118,6 +118,8 @@ func ({{$action.Short}} *{{ $.Pascal }}{{$action.Pascal}}Action) handler(httpReq
 		{{- if $action.RespondJSON }}
 		{{- if $action.Results.Result }}
 		JSON: response.JSON({{ $action.Results.Result }}),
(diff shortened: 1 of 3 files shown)


matthewmueller:
This looks great @theEyeD! Do you mind adding a test for "Allow void view action by passing empty map[string]interface{}{} to view Handler"? 

Then I think this one is good to go!

matthewmueller:
Hey @theEyeD, let me know if you don't have time to wrap this up. I don't want this to get stale.

theEyeD:
@matthewmueller got the original test passing. It required adding `bud.NodeModules["svelte"] = version.Svelte`, which I saw from some other test examples, but I'm not 100% why that's needed. Let me know if that isn't what you were thinking.

matthewmueller:
Excellent thank you Alex! 

The Svelte version is needed because it starts off with a fresh node_modules directory every test run and this test depends on Svelte.

PR #70, opened 2022-05-19
Decorate hot reload welcome message

012e:
I made a welcome message for `bud run`, kinda bland but better than the previous one

### DIFF
--- a/internal/command/run/run.go
+++ b/internal/command/run/run.go
@@ -3,9 +3,11 @@ package run
 import (
 	"context"
 	"net"
+	"time"
 
 	"github.com/livebud/bud/internal/bud"
 	"github.com/livebud/bud/internal/command"
+	"github.com/livebud/bud/internal/version"
 	"github.com/livebud/bud/package/log/console"
 	"github.com/livebud/bud/package/socket"
 )
@@ -19,7 +21,35 @@ type Command struct {
 	Port string
 }
 
+func frontEnd() string {
+	// TODO: Change to the corresponding front end in next releases
+	return "Svelte " + version.Svelte
+}
+
+func displayDashboard(host, port string, timeElapsed time.Duration) {
+	/*
+		The dashboard should looks something like this:
+			|   bud dev server is running:
+			|
+			| > Listening on: http://127.0.0.1:3000
+			| > Front end: Svelte 3.47.0
+			|
+			|   Ready in 270.131758ms
+			|
+	*/
+	address := "http://" + host + ":" + port
+	console.Info("  bud dev server is running:")
+	console.Info("")
+	console.Info("> Listening on: " + address)
+	console.Info("> Front end: " + frontEnd())
+	console.Info("")
+	console.Info("  Ready in " + timeElapsed.String())
+	console.Info("")
+}
+
 func (c *Command) Run(ctx context.Context) error {
+	// Start timer
+	startTime := time.Now()
 	// Start listening on the port
 	listener, err := socket.Load(c.Port)
 	if err != nil {
@@ -34,7 +64,6 @@ func (c *Command) Run(ctx context.Context) error {
 	if host == "::" {
 		host = "0.0.0.0"
 	}
-	console.Info("Listening on http://" + host + ":" + port)
 	// Load the compiler
 	compiler, err := bud.Find(c.bud.Dir)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


matthewmueller:
Thanks for giving this a go @012e! I like the usage of elapsed time, but I'd prefer to keep the prompt fairly minimal. 

I'd suggest:

For boot:

```
Listening on 127.0.0.1:3000
Ready in 300ms
```

Then the "Ready in" line gets replaced each time for successful builds.

1. When a file changes, "Ready in ${n}ms" gets replaced by `Reloading...`
2. Once the build is successful, `Reloading...` gets replaced by `Ready in ${n}ms`
3. If the build is unsuccessful, the error replaces `Reloading...`
4. When a file changes, "Reloading..." is shown again and the cycle repeats.

Optionally, it does the `x3` stuff that Vite does:

```
Ready in 48ms (x2)
```

Is this something you'd be open to trying to get working?

internal/command/run/run.go
```diff
@@ -19,7 +21,35 @@ type Command struct {
...
 
+func frontEnd() string {
+	// TODO: Change to the corresponding front end in next releases
+	return "Svelte " + version.Svelte
+}
+
+func displayDashboard(host, port string, timeElapsed time.Duration) {
+	/*
+		The dashboard should looks something like this:
+			|   bud dev server is running:
```
matthewmueller, line 32: I'd prefer to exclude this information since I think you'll learn that you're running the dev server fairly quickly and never forget it.

internal/command/run/run.go
```diff
@@ -19,7 +21,35 @@ type Command struct {
...
+			|
+			| > Listening on: http://127.0.0.1:3000
+			| > Front end: Svelte 3.47.0
```
matthewmueller, line 35: I'd also exclude the Svelte version. The question I'm asking myself is:

"Does the developer need to make a decision with this information right now?"

I'm concluding no at the moment, but could be convinced otherwise.

internal/command/run/run.go
```diff
@@ -19,7 +21,35 @@ type Command struct {
...
+			|
+			|   Ready in 270.131758ms
```
matthewmueller, line 37: Let's round the elapse time to the nearest millisecond

internal/command/run/run.go
```diff
@@ -50,5 +79,8 @@ func (c *Command) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	// Measure elapsed time
+	timeElapsed := time.Since(startTime)
+	displayDashboard(host, port, timeElapsed)
```
matthewmueller, line 84: Seems to be missing elapsed time on live reload.

matthewmueller (changes requested):

matthewmueller:
https://github.com/livebud/bud/blob/db2d27ba11d3eb8f322f440c9ead74fdb116d9db/runtime/command/run/run.go#L101

matthewmueller:
Hi @012e, thanks for looking into this!

Is it consistently printing out 4 times? If so, that sounds like a bug.

Otherwise, could it be that 4 files changed? In general, the watcher can be slightly flaky (all filewatchers seem to be). You can see what's happening by logging in the watcher itself:

https://github.com/livebud/bud/blob/8e3deacc8f1bbfbf64134214c2e7d06306bda955/package/watcher/watcher.go#L143

012e:
It looks like we are getting the same issue [here](https://github.com/fsnotify/fsnotify/issues/344#issue-679037861), which is caused by calling multiple system events.

I tried to print out the file changed from the `watcher` (budd is just an alias to my build file):
image

Currently, there isn't any official fix for this problem, we will [have to wait](https://github.com/fsnotify/fsnotify/discussions/392#discussion-3535991) for something like `IN_CLOSE_WRITE` API to deal with this.

Any idea, @matthewmueller?

matthewmueller:
Are you on Linux @012e? If so, it sounds like the first step in this task is to create a more reliable watcher. 

How I'd approach this:
1. Verify that you're always getting multiple events for a single file save in the watcher. It seems like you already did this?
2. Try combining multiple events on the same file into one by having a "cool off" period. Something like:
   ```go
   if recentEvents[event.Path].Since(time.Now) > 50*time.Millisecond {
     // ignore 
     continue
   } else {
     recentEvents[event.Path] = time.Now()
   }
   ```

3. If the above fails, create a linux-compatible watcher that has the same API as the existing watcher using build tags or file paths (e.g. `watcher_linux.go`). If fsnotify doesn't work, we could also try: https://github.com/rjeczalik/notify. I tried it early on in the cycle, but it didn't feel quite as robust on OSX. This may have changed or we could have it be the dep for linux.

012e:
OK I will try these out tomorrow

012e:
I made a PR  #77.

PR #75, opened 2022-05-20
Add new/create/edit/update/delete controller verbs

fergloragain:


matthewmueller:
Thanks for giving this a go @fergloragain!

The code you wrote looks good, but I'd like to add the views as well. Would you be interested in contributing the svelte views as well? 

Maybe something like (untested):

**new.svelte**

```
<h1>New {{$.Name.Singular}}</h1>

<form method="post" action="{{ $.Route }}">
  <!-- Add input fields here -->
  <input type="submit" value="Create {{ $.Name }}" />
</form>

<a href="..">Back</a>
```


**edit.svelte**

```
<script>
  export let {{$.Variable}} = {}
</script>

<h1>Edit {{$.Name.Singular}}</h1>

<form method="post" action="{{ $.Route }}">
  <!-- Add input fields here -->
  <input type="hidden" value="{{$.ID}}" />
  <input type="submit" value="Create {{ $.Name }}" />
</form>

<a href="..">Back</a>
```

---

After looking at your PR, I realized that we can't really infer too much about the new and edit views, so I may eventually simplify this command to just be controllers and replace the current command with something like: `bud new resource story:/ title:string url:string body:text`, but for now I think we should align with the index and show actions.

TechplexEngineer:
This is exciting! Have you considered a way for users to add their own templates at a project scope level? 

That was a killer feature of ASP.Net MVC eons ago, we wrote some pretty complex templates and generated 95% of our application.

matthewmueller:
@TechplexEngineer Yep! I'm still working on exposing this, but you'll be able to create your own commands soon, so you could create a `bud new auth` by defining a commands in the project's `./command/new` package:

Something like this:

```go
package new

type Controller struct {
  *scaffold.Template // some scaffolding helper library could be nice
}

func (c *Controller) Auth() error {
  c.Template.Files["controller/admin/admin.go"] = `... some template ...`
  return c.Write()
}
```

Largely inspired by [Laravel's custom command](https://laravel.com/docs/9.x/artisan) support. Anything you can point me to in the ASP.Net MVC world I could have a look at?

matthewmueller:
What do you think about wrapping this one up @fergloragain? Otherwise, just let us know :)

matthewmueller:
@fergloragain re-opening since we definitely want to incorporate these changes! I'll pull this change in and add the views.

PR #77, opened 2022-05-21
Improve hot reload

012e:
### New front end

- Welcome dashboard:

```
Listening on http://127.0.0.1:3000
Ready in 300ms
```

- Better way of prompting successful reloads:

```
Ready in 144ms (x11)
```
### Fix [multiple reload messages](https://github.com/livebud/bud/pull/70#issuecomment-1132497113)

### DIFF
--- a/internal/command/run/run.go
+++ b/internal/command/run/run.go
@@ -2,7 +2,9 @@ package run
 
 import (
 	"context"
+	"fmt"
 	"net"
+	"time"
 
 	"github.com/livebud/bud/internal/bud"
 	"github.com/livebud/bud/internal/command"
@@ -19,7 +21,19 @@ type Command struct {
 	Port string
 }
 
+func displayDashboard(host, port string, timeElapsed int) {
+	/*
+		The dashboard should looks like this:
+			| Listening on: http://127.0.0.1:3000
+			| Ready in 270ms
+	*/
+	console.Info(fmt.Sprintf("Listening on: http://%s:%s", host, port))
+	console.Info(fmt.Sprintf("Ready in %dms", timeElapsed))
+}
+
 func (c *Command) Run(ctx context.Context) error {
+	// Start timer
+	startTime := time.Now()
 	// Start listening on the port
 	listener, err := socket.Load(c.Port)
 	if err != nil {
@@ -34,13 +48,12 @@ func (c *Command) Run(ctx context.Context) error {
 	if host == "::" {
 		host = "0.0.0.0"
 	}
-	console.Info("Listening on http://" + host + ":" + port)
 	// Load the compiler
 	compiler, err := bud.Find(c.bud.Dir)
 	if err != nil {
 		return err
 	}
-	// Compiler the project CLI
+	// Compile the project CLI
 	project, err := compiler.Compile(ctx, &c.bud.Flag)
 	if err != nil {
 		return err
@@ -50,5 +63,8 @@ func (c *Command) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	// Measure elapsed time
+	timeElapsed := int(time.Since(startTime).Milliseconds())
+	displayDashboard(host, port, timeElapsed)
 	return process.Wait()
 }
--- a/package/watcher/watcher.go
+++ b/package/watcher/watcher.go
@@ -6,6 +6,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"time"
 
 	"golang.org/x/sync/errgroup"
 
@@ -133,6 +134,8 @@ func Watch(ctx context.Context, dir string, fn func(path string) error) error {
 	// Note: The FAQ currently says it needs to be in a separate Go routine
 	// https://github.com/fsnotify/fsnotify#faq, so we'll do that.
 	eg, ctx := errgroup.WithContext(ctx)
+	// Start timer
+	startTime := time.Now()
 	eg.Go(func() error {
 		for {
 			select {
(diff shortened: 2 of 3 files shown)


matthewmueller:
Amazing! I'll pull this down today and play with it. Thanks for working on this @012e!!!

matthewmueller:
Yes, I'd try to reduce it as far as possible, while still being reliable enough. Also feel free to bump up the timeout in the test suite if that's the cause of the flakiness..

runtime/command/run/run.go
```diff
@@ -49,56 +52,95 @@ func (c *Command) compileAndStart(ctx context.Context, ln net.Listener) (*exe.Cm
...
+	// Move cursor up and delete line based on ANSI Escape Sequences: https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797
+	switch runtime.GOOS {
+	case "windows":
+		// TODO: Test this on Windows, don't know will it work or not
+		fmt.Print("\033[1A\033[0K")
+	case "linux":
+		fmt.Print("\033[1A\033[0K")
+	default:
+		console.Error("Can't detect your operating system")
+	}
```
matthewmueller, line 65: Let's remove this check.

`fmt.Print("\033[1A\033[0K")` seems to work on OSX and I guess it works for you on Linux. 

Windows experts can help us make this better, when they give Bud a go.

internal/command/run/run.go
```diff
@@ -19,7 +21,19 @@ type Command struct {
 	Port string
 }
 
+func displayDashboard(host, port string, timeElapsed int) {
+	/*
+		The dashboard should looks like this:
+			| Listening on: http://127.0.0.1:3000
+			| Ready in 270ms
+	*/
+	console.Info(fmt.Sprintf("Listening on: http://%s:%s", host, port))
```
matthewmueller, line 30: Remove colon

package/watcher/watcher.go
```diff
@@ -162,11 +165,17 @@ func Watch(ctx context.Context, dir string, fn func(path string) error) error {
 					}
 
 				// Handle write events
-				case op&fsnotify.Write != 0:
+				// Currently, there is an issue with fsnotify which may register multiple events
+				// in a single change, so we have to ignore events that is triggered too
+				// closely together (details in #70)
+				// If there are multiple events in one file change, this only accept the first one.
+				case op&fsnotify.Write != 0 && time.Since(startTime).Milliseconds() > 1:
```
matthewmueller, line 172: This doesn't work on OSX, I guess because events occur soon than 1ms. I think the solution is closer to keeping a map of "last seen" files. If you last saw the file less than a millisecond ago, ignore it.

It's not perfect because that map will keep growing. Maybe we can use an LRU like https://github.com/hashicorp/golang-lru to make it bounded.

runtime/command/run/run.go
```diff
@@ -49,56 +52,95 @@ func (c *Command) compileAndStart(ctx context.Context, ln net.Listener) (*exe.Cm
...
 			}
-			// TODO: host should be dynamic
-			console.Info("Ready on http://127.0.0.1" + c.Port)
+
+			// Time elapsed response
+			timeElapsed := time.Since(startTime).Milliseconds()
+			replacePreviousLine(fmt.Sprintf("Ready in %dms (x%d)", timeElapsed, totalReload))
 			return nil
 		// Hot reload the page
 		default:
```
matthewmueller, line 146: The reloading message should also be shown for non-Go files (e.g. Svelte files) in the `default` case.

matthewmueller (changes requested):
Just reviewed. Nice work Huy! I was able to get this PR working on my machine and it's starting to look pretty great!

One thing I find a bit weird while testing is if you have a `fmt.Println` within your controller, it stops erasing and moves the reload to the new line, so you end up with something like this:

```
| Listening on: http://127.0.0.1:3000
| Ready in 252ms (x1)
| Ready in 261ms (x3)
| Ready in 283ms (x5)
| Ready in 555ms (x12)
| Ready in 470ms (x14)
| Ready in 251ms (x16)
| Ready in 1476ms (x32)
| Ready in 248ms (x33)
| Ready in 250ms (x34)
| Ready in 378ms (x35)
| Ready in 603ms (x36)
| Ready in 438ms (x40)
OK
OK
| Ready in 4671ms (x41)
| Ready in 487ms (x42)
OK....
```

I noticed that Vite "soft clears" upon each reload, where you can still scroll up. Maybe that would improve this behavior? There will need to be a bit of experimentation involved to figure out what feels right.

runtime/command/run/run.go
matthewmueller: If helpful I found [another implementation for a spinner](https://github.com/briandowns/spinner/blob/master/spinner.go#L412-L431), which is used within the [Github's CLI](https://github.com/cli/cli/blob/7a3d02df470252be0dbf8c7bd44176bcd48ba4ef/pkg/iostreams/iostreams.go#L262).

It seems like this one is slightly more battle-tested 😆

package/watcher/watcher.go
matthewmueller: The OSX test suite did correctly report this failure, so you can use that to test your solution.

012e: The problem here is time from this line of code
```golang
// Start timer
startTime := time.Now()
```
to the write handler is too short.
I made [a PR](https://github.com/livebud/bud/pull/88) to for the fix.

012e: The problem here is time from this line of code
```golang
// Start timer
startTime := time.Now()
```
to the write handler is too short.
I made [a PR](https://github.com/livebud/bud/pull/88) to for the fix.

PR #80, opened 2022-05-22
support for Windows builds in make install step

syke99:
This PR adds support for the install step on Windows for preparation of supporting Windows development.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ precommit: test.dev
 install:
 	go mod tidy
 	npm install
-# Note: We don't use a subshell here to support Windows. See #79.
+	@ # Note: We don't use a subshell here to support Windows. See #79.
 	cd livebud
 	npm install
 	cd ..
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ precommit: test.dev
 install:
 	go mod tidy
 	npm install
+# Note: We don't use a subshell here to support Windows. See #79.
 	cd livebud
 	npm install
 	cd ..
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,13 @@ precommit: test.dev
 install:
 	go mod tidy
 	npm install
+ifeq ($(OS), Windows_NT)
+	cd livebud
+	npm install
+	cd ..
+else
 	(cd livebud && npm install)
+endif
 
 ##
 # Examples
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -9,7 +9,13 @@ precommit: test.dev
 install:
 	go mod tidy
 	npm install
+ifeq ($(OS), Windows_NT)
+	cd livebud
+	npm install
+	cd ..
+else
 	(cd livebud && npm install)
+endif
```
matthewmueller, line 18:
```suggestion
	@ # Note: We don't use a subshell here to support Windows. See #79.
	cd livebud
	npm install
	cd ..
```

I'd like to avoid conditionally checking your OS if we can. Since the windows version also works on linux and mac, let's just use that. I also just added a note to my future self not to change it back 😅

matthewmueller (changes requested):

Makefile
syke99: Definitely understandable!! I just pushed up that change 😋

matthewmueller: Thanks @syke99! By the way, next time you can just hit the "Commit suggestion" button if the change works for you 😌

matthewmueller:
For some reason this change contains https://github.com/livebud/bud/commit/a6a8b1202ceff372771e5b50376d3cc3261918a3

Could you maybe rebase and push back up?

Makefile
```diff
@@ -9,7 +9,10 @@ precommit: test.dev
 install:
 	go mod tidy
 	npm install
-	(cd livebud && npm install)
+# Note: We don't use a subshell here to support Windows. See #79.
```
matthewmueller, line 12: This should be tabbed and be prefixed with `@ ` so it doesn't show up when you run `make install`

matthewmueller (changes requested):

Makefile
syke99: Ahh, didn't even see that option until now 😂 Thanks, I'll use it next time 😋

syke99:
Okay, repo has been rebased, comment corrected in Makefile, and changes pushed up~~ 😋

matthewmueller:
Hey @syke99, still seeing a bunch of files being updated. Maybe close this PR and open a new one?

syke99:
Sounds good, not sure why those commits are tagged to this PR. But I'll get a new one out!!

Issue #82, opened 2022-05-22
Nested view URL don't prefix the URL with the nested route


iAziz786:
I created nested views with running the command.

```sh
bud new controller post index show
```

It generated the files fine but when I opened the `/post` URL, the `td` element didn't include prefix the URL with the `/post` instead it relative to the root.

<img width="1450" alt="image" src="https://user-images.githubusercontent.com/17024120/169707563-cbe3d853-53be-47f7-b155-46966da05632.png">

<img width="1494" alt="image" src="https://user-images.githubusercontent.com/17024120/169707511-a7fef1e7-913a-4dc0-bd04-009fa4d898be.png">

I'm not sure whether I was suppose to update the URL manually or not as it's a minor thing.

matthewmueller:
Good catch! Just opened a PR: https://github.com/livebud/bud/pull/84

PR #84, opened 2022-05-23
runtime/command/new/controller: Fix links when generating controllers and views

matthewmueller:
Closes: #82

### DIFF
--- a/runtime/command/new/controller/view/index.gotext
+++ b/runtime/command/new/controller/view/index.gotext
@@ -15,8 +15,8 @@
   {#each {{ $.Plural }} as {{ $.Singular -}} }
     <tr>
       {#each Object.keys({{ $.Singular }}) as key}
-        {#if key === "id"}
-          <td><a href={`/${ {{- $.Singular }}[key]}`}>{ {{- $.Singular }}[key]}</a></td>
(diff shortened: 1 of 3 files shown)


runtime/command/new/controller/view/index.gotext
```diff
@@ -15,8 +15,8 @@
   {#each {{ $.Plural }} as {{ $.Singular -}} }
     <tr>
       {#each Object.keys({{ $.Singular }}) as key}
-        {#if key === "id"}
```
matthewmueller, line 18: Minor tweak so if you haven't specified `json:"id"`, your IDs will still work.

I think we can go even further here and say whatever the first key is, we'll assume it's your primary key. 

Most people will end up deleting this anyway, so whatever we can do to help people get up and running quicker, the better.

runtime/command/new/controller/view/show.gotext
```diff
@@ -19,7 +19,7 @@
 
 <br />
 
-<a href="..">Back</a>
```
matthewmueller, line 22: I swore this worked in my testing, but I guess not. `/post/1` would incorrectly resolve to `/` not `/post`.

PR #85, opened 2022-05-23
new fork to exclude external commits

syke99:
This should remove the commits to other files being tied to the PR like the previous PR for this fix was experiencing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,10 @@ precommit: test.dev
 install:
 	go mod tidy
 	npm install
-	(cd livebud && npm install)
+	@ # Note: We don't use a subshell here to support Windows. See #79.
+	cd livebud
+	npm install
+	cd ..
 
 ##
 # Examples
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,10 @@
 BUD_VERSION := $(shell cat version.txt)
 
+## Variables for getting current directory to 
+#  allow for Windows support
+mkfile_path := $(abspath $(lastword $(MAKEFILE_LIST)))
+current_dir := $(notdir $(patsubst %/,%,$(dir $(mkfile_path))))
+
 precommit: test.dev
 
 ##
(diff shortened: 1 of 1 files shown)


syke99:
This PR has been converted to a draft as I need to continue getting the example Make commands to work with Windows. (The npm link command is failing to find the package.json. It's malforming the path, so may need to remove Make with Chocolatey and re-install another way)

matthewmueller:
Hey @syke99, thank you for the effort. I must admit, the more I reflect on these changes, the more concerned I get.

We're making the Makefile more fragile for our current target users (OSX and Linux) while trying to adapt it for a platform we currently don't support. We still have some large hills to climb in order to properly support Windows (V8 support & passing file descriptors between binaries).

As I look to Laravel, I see that they also recommend using Laravel with WSL2 and it's not clear that they support Windows native. We're far from the capabilities of Laravel and I'd like to focus our effort on catching up before expanding platforms.

What do you think? I'd still like to support native Windows at some point, it's one of the key benefits of using Go, I just don't think we have enough usage or expertise yet to add proper support for the platform.

If the above makes sense, I have some suggestions:
- I'm open to continuing this effort if we can properly abstract these concepts in a cross-compatible way. I like our approach with `mkfile_path`, could we have windows-compatible functions for spawning subshells?
- Otherwise, we add a note in the Contributing Guide for Windows users that we recommend using WSL2

syke99:
Hey Matt, I’m 100% okay with just adding a note in the README for users to just use WSL2. Honestly, getting native Windows development to work is going to be a HEAVY lift, especially with the v8 and the file descriptors/handlers. It’s something I’d still like to tinker with, but I’m definitely alright with not putting nearly as much focus on it right now!!

matthewmueller:
Sounds good, thanks @syke99. Closing this PR then, but I'd be open to adding a note in the contributor guide for native Windows users!

PR #88, opened 2022-05-25
Fix multiple reloads in one change

012e:
Simply [fix multiple reloads in one change](https://github.com/livebud/bud/pull/70#issuecomment-1132497113)


matthewmueller:
Hi @012e thanks for your effort, but I don't think this solution is going to work. 

- It might work on the first reload, but not subsequent reloads
- If you subtract 1ms every reload, you're back at where it is right now

I'd recommend trying the solution I suggested earlier. 

By the way, I'm a bit confused why the Ubuntu tests are passing the watcher, but the watcher isn't working for your machine. Do you have any idea why that might be?

012e:
This code only initialize the `startTime` variable once, minus by 1ms to ensure it won't get ignore by `time.Since(startTime).Milliseconds() > 1`.
```golang
startTime := time.Now().Add(-1 * time.Millisecond)
```

Each time an event is triggered `startTtime` will be reset.
```golang
startTime = time.Now()
```
image

For example, if 2 events are triggered in one change, 0ms between each. Then time between (3) to (4) is also 0ms and ignored.

I tested it on my local machine, it works properly, there's no extra reloads. I did test [CI/CD twice on my fork](https://github.com/012e/bud/actions/runs/2382466161) before pushing. It works.

Now I am also confused 😕.

012e:
> By the way, I'm a bit confused why the Ubuntu tests are passing the watcher, but the watcher isn't working for your machine. Do you have any idea why that might be?

1. Probably because it changes the file one time, and `getEvent` only expect *at least* a response to pass the test. So if there are more, the test still doesn't fail.
https://github.com/livebud/bud/blob/ee97bd4d1baceaa45feaaa612f1619985c29ea1f/package/watcher/watcher_test.go#L35-L41
2. Is it something related to system hardware or drivers + OS?

matthewmueller:
Hmm, okay let me pull this done and test it locally! Cool diagram btw :-D

package/watcher/watcher.go
```diff
@@ -162,11 +165,17 @@ func Watch(ctx context.Context, dir string, fn func(path string) error) error {
 					}
 
 				// Handle write events
-				case op&fsnotify.Write != 0:
+				// Currently, there is an issue with fsnotify which may register multiple events
+				// in a single change, so we have to ignore events that is triggered too
+				// closely together (details in #70)
+				// If there are multiple events in one file change, this only accept the first one.
+				case op&fsnotify.Write != 0 && time.Since(startTime).Milliseconds() >= 1:
+					startTime = time.Now()
```
012e, line 173: My mistake, the code just ignored *all events* that is triggered within 1ms instead of just *write events*

012e:
I think this will work 😀

matthewmueller:
Hey @012e, just pulled it down. It seems to be working. Can you remind me of the problem this solves for you? From the previous PR it was to avoid multiple triggers, is that still the problem? 

With this PR, I'm still a bit hesitant: if a single new file change occurs within 1 millisecond of the last file change, the last file change is lost, leading `bud/` build to be older than what's in your filesystem. This is hypothetical though since I'm not able to reproduce this behavior.

012e:
### My issue: 

It reloads multiple times because of multiple write signals.
Normally, when I make a *single* file change while `bud run`:
```
| Ready on http://127.0.0.1:3000
| Ready on http://127.0.0.1:3000
| Ready on http://127.0.0.1:3000
| Ready on http://127.0.0.1:3000
``` 
- If I add this line here ([fix-multiple-reloads](https://github.com/012e/bud/tree/fix-multiple-reloads) branch)

```go
case evt := <-watcher.Events:
        // Add here
         fmt.Printf("Current event: %s, time between previous event: %s\n", evt, time.Since(startTime).String())
	switch op := evt.Op; {
```

- Each time I make a *single* file save:

```
Current event: "controller/controller.go": WRITE, time between previous event: 1.407399441s
| Ready on http://127.0.0.1:3000
Current event: "controller/controller.go": WRITE, time between previous event: 2.25µs
Current event: "controller/controller.go": WRITE, time between previous event: 27.773µs
Current event: "controller/controller.go": WRITE, time between previous event: 33.173µs
```
### bud run

> if a single new file change occurs within 1 millisecond of the last file change, the last file change is lost

OK, because the write handler wasn't triggered (event triggered within 1ms cool down period)

> leading `bud/` build to be older than what's in your filesystem

Uh, since the watcher didn't emit a **write** event, no re-compile is made. This is the flaw using a *cool down*, not much way to prevent.

matthewmueller:
Just tested again, you're not seeing 2 events per save on your linux? I created a test script:

**scripts/test-watcher/main.go**

```
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/livebud/bud/internal/current"

	"github.com/livebud/bud/internal/sig"
	"github.com/livebud/bud/package/watcher"
)

func run(ctx context.Context) error {
	dirname, err := current.Directory()
	if err != nil {
		return err
	}
	ctx = sig.Trap(ctx, os.Interrupt)
	return watcher.Watch(ctx, dirname, func(path string) error {
		fmt.Println("-> triggered", path)
		return nil
	})
}

func main() {
	ctx := context.Background()
	if err := run(ctx); err != nil {
		log.Fatal(err)
	}
}
```

**scripts/test-watcher/a.txt**

```
aa
```

Then using the Ubuntu image in the [contributing](./contributing) Readme

```
docker build -t bud:latest contributing
docker run -it --rm -v $(PWD):/bud bud /bin/bash
```

Results:

```
docker run -it --rm -v $(PWD):/bud bud /bin/bash
root@5d9c4cac4118:/bud# go run scripts/test-watcher/main.go 
-> triggered /bud/scripts/test-watcher/a.txt
-> triggered /bud/scripts/test-watcher/a.txt
```

Whenever I make a change, it triggers twice.

PR #93, opened 2022-05-28
Refactor and test the compiler

matthewmueller:
This is a mammoth PR that I should have broken up but... `¯\_(ツ)_/¯`

The focus of this PR was to:
1. Add a lot more testing to key pieces of the framework. 
2. Start to simplify and reduce the number of packages with the same name. 
3. Double down on the existing compiler architecture.
4. Prepare the framework for upcoming features.

User facing changes:
- `bud run --port` renamed to `bud run --listen`
- Exit codes may have changed (to be tested further)


internal/generator/command/command.gotext
```diff
@@ -10,48 +10,82 @@ import (
...
 	{ // cli run
-		cmd := &run.Command{Flag: c.flag, Project: project}
 		cli := cli.Command("run", "run command")
-		cli.Flag("port", "port to listen to").String(&cmd.Port).Default(":3000")
-		cli.Run(cmd.Run)
+		cli.Flag("embed", "embed assets").Bool(&c.runCmd.Flag.Embed).Default(false)
+		cli.Flag("hot", "hot reloading").String(&c.runCmd.Flag.Hot).Default(":35729")
+		cli.Flag("minify", "minify assets").Bool(&c.runCmd.Flag.Minify).Default(false)
+		cli.Flag("listen", "address to listen to").String(&c.runCmd.Listen).Default(":3000")
+		cli.Run(c.beforeRun(c.runCmd))
```
matthewmueller, line 42: Moves more of the core logic into the project cli generator

internal/generator/command/parser.go
```diff
@@ -92,6 +92,35 @@ func (p *parser) loadCommand2(parent *Cmd, base, dir string) *Cmd {
 	return cmd
 }
 
+func (p *parser) loadProvider() *di.Provider {
```
matthewmueller, line 95: Added generator dependency injection to the generated command package, so commands like run and build will have the dependencies they need to build the application generators.

matthewmueller: This is in addition to dependency injection by the project cli to support custom dependencies in custom cli commands.

internal/is/is.go
```diff
@@ -0,0 +1,124 @@
+// Package is has a modified superset of the excellent github.com/matryer/is.
+//
+// It adds new methods commonly needed in Bud, removes indentation and replaces
+// displaying comments with optional messages that can be formatted.
+//
+// Learn more in this discussion: https://github.com/livebud/bud/discussions/86
```
matthewmueller, line 6: Internalized matryer/is

internal/once/error.go
```diff
@@ -0,0 +1,13 @@
...
+
+type Error struct {
+	o   sync.Once
+	err error
+}
+
+func (e *Error) Do(fn func() error) (err error) {
+	e.o.Do(func() { e.err = fn() })
+	return e.err
+}
```
matthewmueller, line 13: Simple sync.Once helper package

internal/sig/trap.go
```diff
@@ -7,14 +7,14 @@ import (
...
 	ch := make(chan os.Signal, len(signals))
 	go func() {
 		<-ch
-		cancel()
 		signal.Stop(ch)
+		cancel()
 	}()
 	signal.Notify(ch, signals...)
-	return ret, cancel
+	return ret
```
matthewmueller, line 19: This was done because if a command finished (got to the end of the `Run(ctx) error` function), its context would be canceled. 

This was surprising behavior since this package was primarily just meant to cancel upon a signal. To reduce confusion, this PR removes the ability to cancel a trap manually. 

You could still wrap the trap ctx in a cancellable context if you want the old behavior

internal/snapshot/snapshot_test.go
```diff
@@ -92,3 +92,6 @@ func TestRestoreNotExist(t *testing.T) {
 	is.True(errors.Is(err, fs.ErrNotExist))
 	is.Equal(fsys, nil)
 }
+
+// TODO: test changed files within replaced modules
+// triggering hash changes
```
matthewmueller, line 97: Missing test

internal/testdir/testdir.go
```diff
@@ -42,8 +38,11 @@ const goMod = `
 	)
 `
 
-func New() *Dir {
+func New(dir string) *Dir {
 	return &Dir{
+		dir:         dir,
+		Backup:      false, // TODO: re-enable and store snapshots in ./bud/tmp
```
matthewmueller, line 44: Snapshotting and restoring test directories have been turned off by default. I want to re-enable this again when there's more tests (like the one above) and it's hooked into CI.

internal/testdir/testdir.go
```diff
@@ -211,15 +200,39 @@ func (d *Dir) Hash() (string, error) {
 	return snapshot.Hash(fsys)
 }
 
+func (d *Dir) writeAll(fsys fs.FS, to string) error {
```
matthewmueller, line 203: testdir no longer "syncs" files (where it removes files it doesn't know about), instead multiple testdir.Write() will now override files in its list.

package/commander/arg.go
```diff
@@ -17,18 +17,20 @@ func (a *Arg) String(target *string) *String {
...
 	a.value = &stringMapValue{inner: value}
 	return value
 }
 
+// Custom allows you to define a custom parsing function
+func (a *Arg) Custom(fn func(string) error) *Custom {
+	value := &Custom{target: fn}
+	a.value = &customValue{inner: value}
+	return value
+}
```
matthewmueller, line 32: Add custom flag and argument parsing (currently unused), but may be supported in the future. Similar to the standard library's `flag.Func`

package/exe/exe.go
```diff
@@ -62,19 +76,25 @@ func isWaitError(err error) bool {
 }
 
 func (c *Cmd) Run() error {
-	return c.cmd().Run()
+	if err := (*exec.Cmd)(c).Run(); err != nil {
+		if canIgnore(err) {
```
matthewmueller, line 80: Tried to reduce the duplicate `signal: killed` logging. This will likely need to be revisited because I don't think the correct exit codes are being returned now under certain conditions.

package/gomod/mod.go
```diff
@@ -102,6 +102,9 @@ func parse(opt *option, path string, data []byte) (*Module, error) {
 	if err != nil {
 		return nil, err
 	}
+	if modfile.Module == nil {
+		return nil, fmt.Errorf("mod: missing module statement in %q", path)
+	}
```
matthewmueller, line 107: Fixes periodic panics in CI. Still trying to get to the bottom of why this sometimes occurs only in CI.

matthewmueller: This happens when the go.mod exists but doesn't have a `module <name>` statement

package/hot/client.go
```diff
@@ -0,0 +1,167 @@
+package hot
```
matthewmueller, line 1: Added a server-sent events (SSE) client to test hot reloads better!

package/log/log.go
```diff
@@ -42,7 +42,7 @@ type Flusher interface {
 	Flush()
 }
 
-type Logger interface {
+type Interface interface {
```
matthewmueller, line 45: log.Logger would always default to the standard libary's log package. This fixes that.

package/merged/merged.go
```diff
@@ -47,9 +53,32 @@ func (f *FS) Open(path string) (fs.File, error) {
...
+// Error implements error and joins all the errors together
+func (n *notExists) Error() string {
+	return fmt.Errorf("merged: open %q. %w", n.path, errs.Join(n.errors...)).Error()
+}
+
+// This type of error should be an fs.ErrNotExist because all underlying errors
+// are also fs.ErrNotExists
+func (*notExists) Unwrap() error {
+	return fs.ErrNotExist
+}
```
matthewmueller, line 80: This is a surprisingly important fix. 

The error chain was getting lost when looking up a file that didn't exist in any filesystem (this can happen when there's errors). 

We now collect and continue the chain when we encounter `fs.ErrNotExist` errors. 

The collected errors themselves also satisfy `errors.Is(err, fs.ErrNotExist)`.

runtime/command/run/run.go
```diff
@@ -2,75 +2,132 @@ package run
...
+	if err := eg.Wait(); err != nil {
+		return err
+	}
+	return nil
 }
 
-func (c *Command) compileAndStart(ctx context.Context, ln net.Listener) (*exe.Cmd, error) {
-	app, err := c.Project.Compile(ctx, c.Flag)
-	if err != nil {
+func (c *Command) compileAndStart(ctx context.Context, listener socket.Listener) (*exe.Cmd, error) {
```
matthewmueller, line 63: Moved most of the `bud {run,build,...}` logic into the runtime to prep for wrapping built-in commands.

runtime/command/run/run.go
```diff
@@ -87,18 +144,13 @@ func (c *Command) startApp(ctx context.Context, hotServer *hot.Server) error {
...
-			}
-			process, err = app.Start(ctx, listener)
+			p, err := c.compileAndStart(ctx, listener)
 			if err != nil {
 				console.Error(err.Error())
 				return nil
 			}
-			// TODO: host should be dynamic
-			console.Info("Ready on http://127.0.0.1" + c.Port)
+			process = p
```
matthewmueller, line 152: ^ Fixes a panic where failures would cause the process to be nil, if the watch command exited, process.Close would panic.

runtime/generator/controller/controller_test.go
```diff
@@ -4,159 +4,277 @@ import (
...
 		Content-Type: application/json
-		Date: Fri, 31 Dec 2021 00:00:00 GMT
 
-		"About"
-	`))
+		"Comments"
+	`)
+	// Test stdio
+	is.Equal(stdout.String(), "")
+	is.Equal(stderr.String(), "")
```
matthewmueller, line 102: Lots of tests needed to be refactored to use the new testcli package.

matthewmueller (commented):
Hopefully not such a big PR in the future!

Issue #90, opened 2022-05-27
../../../go/pkg/mod/github.com/livebud/bud@v0.1.4/package/scaffold/template.go:28:8: undefined: any


forging2012:
```
➜  bud-hello bud version
     bud: 0.1.4
  svelte: 3.47.0
   react: 18.0.0
➜  bud-hello bud run
| Listening on http://127.0.0.1:3000
# github.com/livebud/bud/package/scaffold
../../../go/pkg/mod/github.com/livebud/bud@v0.1.4/package/scaffold/template.go:28:8: undefined: any
note: module requires Go 1.18
➜  bud-hello go version
go version go1.17.10 darwin/amd64
```

matthewmueller:
Oh wow, looks like an `any` snuck in there, thanks! I'll setup CI to catch this.

forging2012:
```
➜  bud-hello bud version
     bud: 0.1.4
  svelte: 3.47.0
   react: 18.0.0
➜  bud-hello go version
go version go1.18.2 darwin/amd64
➜  bud-hello bud run
| Listening on http://127.0.0.1:3000
build flag -mod=mod only valid when using modules
| exit status 1
```

forging2012:
```
➜  bud-new go env -w GO111MODULE=on
➜  bud-new bud run
| Listening on http://127.0.0.1:3000
```
It works~!

matthewmueller:
@forging2012 glad to hear you got it working. I'm surprised `GO111MODULE=on` was not the default for `Go 1.18`. Are you explicitly overriding that in your dotfiles or something?

forging2012:
```
I update golang version to go1.18.2.
Error follow:

➜  bud-hello bud version
     bud: 0.1.4
  svelte: 3.47.0
   react: 18.0.0
➜  bud-hello go version
go version go1.18.2 darwin/amd64
➜  bud-hello bud run
| Listening on http://127.0.0.1:3000
build flag -mod=mod only valid when using modules
| exit status 1
```

When I run:
go env -w GO111MODULE=on
bud run
| Listening on http://127.0.0.1:3000
It works for me ~!

matthewmueller:
Hmm, I'm not able to reproduce the `GO111MODULE=on` requirement. A couple questions:

1. What do you get when you run `go env GO111MODULE`?
2. What does `ls hello-bud` show?

forging2012:
```
1、➜  bud-hello go env GO111MODULE
on
2、
[...]

PR #95, opened 2022-05-28
ci: add matrix testing for supported Go and Node.js versions

wheinze:
Also,
- ci: do not fail fast to preserve error messages
- ci: prefer built-in caching for `actions/setup-go`
- chore: update to the latest gh actions

This PR is only meant to add support for matrix testing of supported Go and Node.js versions as proclaimed in https://github.com/livebud/bud/issues/90#issuecomment-1139315192 .

I would propose that back-porting bud to older versions of Go should be done separately. The matrix testing already gives us hints about the lacking language features of older Go versions ;)


.github/workflows/test.yml
```diff
@@ -2,32 +2,50 @@ name: Test
...
 
       - name: Set up Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@v3
         id: go
         with:
-          go-version: 1.18.0
+          go-version: ${{ matrix.go }}
+          cache: true
+          cache-dependency-path: go.sum
```
wheinze, line 43: caching is built-in since v3, so we can omit the previous workaround

.github/workflows/test.yml
```diff
@@ -2,32 +2,50 @@ name: Test
...
+            node: 14
+          - os: ubuntu-latest
+            go: 1.18
+            node: 18
+          - os: macos-latest
+            go: 1.18
+            node: 14
+          - os: macos-latest
+            go: 1.18
+            node: 18
```
wheinze, line 31: only do the full Node.js matrix testing for the latest Go version

matthewmueller: Looks good, thanks!

.github/workflows/test.yml
```diff
@@ -2,32 +2,50 @@ name: Test
 
 on:
   push:
-    branches: ['*']
+    branches: [main]
   pull_request:
-    branches: ['*']
+    branches: [main]
```
matthewmueller, line 7: Does this mean it will only trigger when PRs are opened against main or the main branch is updated?

matthewmueller:
Thanks Waldemar! Improving CI is very much appreciated. 

One other thing I'd like to do (in a separate PR) is to cache [buildcache binaries](https://github.com/livebud/bud/blob/main/internal/buildcache/buildcache.go#L19). This would drastically speed up the tests, so we no longer need to run `go build` per test. 

The existing yaml cache configuration isn't doing anything currently. Do you have any idea how to fix this? If so, PRs welcome!!

.github/workflows/test.yml
wheinze: Yep, and not only on opening new PRs, but also on new pushes to them.  At least that's how I'm used to use this most of the time to reduce load on CI, but if you intended some other cases that aren't matched hereby, I'm happy to revert this change ;)

.github/workflows/test.yml
wheinze: Just noticed that it was part of the recent release of this action while upgrading and was hoping it would work this way straight out of the box. I will have a deeper look on this and play around a little bit with the action. 

If you prefer to merge the matrix testing beforehand, I could easily revert the previous caching mechanism and provide an upcoming PR once I'm confident how to use it via the action itself.

Note: A first look on the action itself seems promising (see: https://github.com/actions/setup-go/blob/b22fbbc2921299758641fab08929b4ac52b32923/src/package-managers.ts#L13) as at least the same directories are used ;)

.github/workflows/test.yml
```diff
@@ -40,38 +58,19 @@ jobs:
...
 
       - name: Cached testdir snapshots
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: $TMPDIR/bud/snapshot
           key: ${{ runner.os }}
 
-      - name: Run CI (OSX)
+      - name: Run CI (Ubuntu Linux)
```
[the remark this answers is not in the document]
wheinze: This was really just a minor consistency adjustment to align the order as ubuntu is also used in the first place throughout this file before.

wheinze:
> Thanks Waldemar! Improving CI is very much appreciated.

My pleasure, will see how I can support here ;)

> One other thing I'd like to do (in a separate PR) is to cache [buildcache binaries](https://github.com/livebud/bud/blob/main/internal/buildcache/buildcache.go#L19). This would drastically speed up the tests, so we no longer need to run `go build` per test.

Sounds great, let's try to do so!

> The existing yaml cache configuration isn't doing anything currently. Do you have any idea how to fix this? If so, PRs welcome!!

My bad, I was just stumbling upon this in their changelog while upgrading to the latest version of the action, but I'll have a deeper look on this (see also: https://github.com/livebud/bud/pull/95#discussion_r884279201).

Another outstanding question from my side is, if there is a need to support Go 1.16, or if we should concentrate on 1.17 and 1.18 like to Go team does it? But this question might not be discussed in this PR.

.github/workflows/test.yml
wheinze: Spawned up a little [playground](https://github.com/wheinze/setup-go-playground/actions) to check the action and actually it seems to work fine for me.

So I would recommend that you give it a try and grant the second actions run to check if the cache will be restored ;) Or do I miss something?

matthewmueller: I think it's fine to merge this as is. 

I didn't spend a lot of time testing or verifying CI yet, so I'm not even sure the existing setup is doing the correct caching. 

Thanks for testing this in a playground!

matthewmueller:
Working on merging this, I'd like to get this passing in CI before merging, so we're not breaking on main. 

There's just one more thing to resolve which is the bud binary $PATH issue that's failing. I've been wanting to fix this and this PR is giving me the motivation to do so.

> Another outstanding question from my side is, if there is a need to support Go 1.16, or if we should concentrate on 1.17 and 1.18 like to Go team does it? But this question might not be discussed in this PR.

Does Go team not support less than 2 versions back? My understanding is that they

matthewmueller:
This is now in main via https://github.com/livebud/bud/pull/118. Thanks @wheinze!

PR #133, opened 2022-06-06
Experimental compiler refactor

matthewmueller:
I'd like to take one more crack at the compiler refactor. It's important to get this process right because all the generators will fit within this overall process. Writing generators is the easy part.

The goal is to:
- Transition from bud starting a long-running generated cli process that starts another long-running app process to bud optionally running a short-lived generated cli process, then running a long-running app process.  
 
Additionally:
- Gather all the generators and runtimes into a simple `framework/` directory

This will hopefully make the code base much easier to follow, without compromising on flexibility.


matthewmueller:
Woohoo, it's aliveeeeee. Just need to do a bit more cleanup and it's good to go!

framework/controller/loader.go
```diff
@@ -264,7 +266,7 @@ func (l *loader) loadView(controllerKey, actionKey, actionRoute string) *View {
 		if base != key {
 			continue
 		}
-		l.imports.Add(l.module.Import("bud/.app/view"))
+		l.imports.Add(l.module.Import("bud/internal/app/view"))
```
matthewmueller, line 269: I thought `.app` was pretty clever since it the dot causes `goimports` to ignore it, but you can still directly import it. 

This did have the unfortunate side effect that `go mod tidy` would remove dependencies that were only present in `.app/*` because it skipped over them. To counteract this, I added a generated `bud/imports.go` that would link into `.app/*`

Going with `internal/` should simplify everything. They aren't `goimport`ed because they're internal to `bud/`, `go mod tidy` should still pick them up because it's within the module in a standard directory. This also allowed us to remove the `bud/imports.go`.

framework/controller/loader.go
```diff
@@ -278,7 +280,7 @@ func (l *loader) loadActionParams(params []*parser.Param) (inputs []*ActionParam
 		inputs = append(inputs, l.loadActionParam(param, nth, numParams))
 	}
 	if len(inputs) > 0 {
-		l.imports.Add("github.com/livebud/bud/runtime/controller/request")
+		l.imports.Add("github.com/livebud/bud/framework/controller/controllerrt/request")
```
matthewmueller, line 283: All helper runtimes live within the generator's package inside the framework directory.

framework/framework.go
```diff
@@ -0,0 +1,8 @@
+package framework
+
+// Flag is used by many of the framework generators
+type Flag struct {
```
matthewmueller, line 4: This might not be the best place for this, but I basically needed configuration I could share to a lot of packages that doesn't have a lot of imports avoiding circular dependencies.

framework/public/loader.go
```diff
@@ -89,14 +87,6 @@ func (l *loader) loadEmbedsFrom(root, dir string) (files []*embed.File) {
...
 }
 
-//go:embed favicon.ico
-var favicon []byte
-
-// Default CSS is modern-normalize by Sindre Sorhus
-// https://raw.githubusercontent.com/sindresorhus/modern-normalize/v1.1.0/modern-normalize.css
-//go:embed default.css
-var defaultCSS []byte
-
```
matthewmueller, line 99: Moved to the view generator, which is a better spot for this.

framework/view/loader.go
```diff
@@ -0,0 +1,98 @@
...
+		domCompiler := dom.New(l.module, l.transform.DOM)
+		files, err := domCompiler.Compile(ctx, l.fsys)
+		if err != nil {
+			fmt.Println("error compiling", err)
+			return nil, err
+		}
+		for _, file := range files {
+			// TODO: decide if we should be doing strings.ToLower here. It's needed
+			// because the router always lower-cases things, but the generated chunks
+			// contain are upper values
```
matthewmueller, line 80: Oh interesting, this should get revisited now that we have #142.

framework/view/ssr/jsx.gotext
```diff
@@ -1,4 +1,4 @@
-import { createView } from "./bud/view/_jsx.ts"
+import { createView } from "./bud/view/_jsx.ts.js"
```
matthewmueller, line 1: Not having `.js` prefix was causing some issues in ecosystem tooling that uses the extension for certain behavior.

framework/view/view.gotext
```diff
@@ -0,0 +1,34 @@
...
+	{{- range $import := $.Imports }}
+	{{$import.Name}} "{{$import.Path}}"
+	{{- end }}
+)
+{{- end }}
+
+{{- if not $.Flag.Embed }}
+// Load the view server. Files are linked rather than embedded.
+func Load(client budclient.Client) Server {
+	return viewrt.Proxy(client)
```
matthewmueller, line 18: Requests for client-side built views are now proxied to the bud server to be built and rendered

internal/cli/bud/bud.go
```diff
@@ -0,0 +1,178 @@
...
+//   b. Generate custom command
+//     i. Generate bud/internal/command/${name}/
+//     ii. Build bud/command/${name}
+// 2. Run bud/command/${name}
+func (c *Command) Run(ctx context.Context) error {
+	return commander.Usage()
+}
+
+// Module finds the go.mod file for the application
+func Module(dir string) (*gomod.Module, error) {
```
matthewmueller, line 76: internal/cli/bud contains some shared configuration, the root command and some helper functions that are used by the rest of the CLI

internal/cli/toolfscat/toolfscat.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+	"io/fs"
+	"os"
+	"path"
+
+	"github.com/livebud/bud/framework"
+	"github.com/livebud/bud/internal/cli/bud"
+)
+
+func New(bud *bud.Command, in *bud.Input) *Command {
```
matthewmueller, line 14: Added some tools to inspect the generated files

internal/cli/toolfstxtar/toolfstxtar.go
```diff
@@ -0,0 +1,68 @@
...
+			Data: code,
+		})
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	// Print the archive to stdout
+	fmt.Fprintln(os.Stdout, string(txtar.Format(ar)))
+	return nil
```
matthewmueller, line 67: This one is going to be handy for reproing issues

internal/exe/exe.go
```diff
@@ -0,0 +1,163 @@
...
+
+import (
+	"context"
+	"errors"
+	"io"
+	"os"
+	"os/exec"
+
+	"github.com/livebud/bud/internal/once"
+)
```
matthewmueller, line 11: This package simplifies the previous exe package to consolidate how subprocesses are managed in bud. Instead of wrapping exec.Command, it can turn an exec.Cmd into a Process.

matthewmueller: Additionally, it can allow you to configure commands earlier with shared env / stdio / files etc. that can be spawned at a later time.

internal/format/format.go
```diff
@@ -0,0 +1,20 @@
...
+
+func format(message string) string {
+	return strings.TrimSpace(dedent.Dedent(message)) + "\n"
+}
+
+func Sprintf(message string, args ...interface{}) string {
+	return fmt.Sprintf(format(message), args...)
+}
+
+func Errorf(message string, args ...interface{}) error {
```
matthewmueller, line 18: Added a way to add multi-line errors more easily.

matthewmueller:
This PR concludes the large internal refactor I've been working on since launch. Since so much of the codebase has changed, I'm pretty sure there's going to be a few minor regressions, but we can address them as they pop up. 

The results of this PR:
- Code is much simpler, there's no more generated CLI, we're just generating an app. 
- All generators and helper runtimes live in `framework/`, supporting packages that are needed by the generated code live in `packages/`, everything else goes in `internal/`.
- Tests are faster and less flaky because no more generated CLI and a pubsub system that will tell the test when the app is ready
- Tests are more end-to-end because they only support two commands: cli.Run and cli.Start. 
- No more v8client/v8server, everything has been consolidated into the bud client/server that's run during `bud run`.
- Debug logging! Added --log=debug to bud and the test suite. Slowly passing the logger through the subsystems.

PR #140, opened 2022-06-18
contributing: update Go and Node.js versions in Dockerfile

wheinze:
Updates the OS and Go to the latest, and Node.js to the latest LTS versions.


contributing/Dockerfile
```diff
@@ -1,27 +1,25 @@
-FROM ubuntu:18.04
+ARG NODE_VERSION=16.15.1
+FROM node:${NODE_VERSION}-slim
```
matthewmueller, line 2: Any advantage of doing it like this? 

I find this a bit harder to understand when we're starting with a Node image and extending it rather than starting with a distro and adding Node.

Questions that pop in my head:
- What distro am I working with? Is it Alpine, Ubuntu, Centos, etc.

contributing/Dockerfile
```diff
@@ -1,27 +1,25 @@
...
 RUN node -v
 
+# Install basic dependencies
+RUN apt-get -qq update \
+  && apt-get -qq -y install wget git make gcc g++ \
+  && rm -rf /var/lib/apt/lists/*
+
 # Install Go
-RUN curl -L https://golang.org/dl/go$GO_VERSION.linux-amd64.tar.gz | tar -xz -C /usr/local --strip-components 1
+RUN wget -q -O - https://go.dev/dl/go${GO_VERSION}.linux-amd64.tar.gz | tar -xz -C /usr/local --strip-components 1
```
matthewmueller, line 15: Any reason for wget over curl?

matthewmueller:
Hey @wheinze, thanks for the PR! 

Could you motivate these changes a bit? Why the current Dockerfile needs improvement, etc.

wheinze:
> Hey @wheinze, thanks for the PR!
> 
> Could you motivate these changes a bit? Why the current Dockerfile needs improvement, etc.

Sure, added some more info ;)

contributing/Dockerfile
matthewmueller: Awesome, thanks so much for the context @wheinze. I had no idea Debian was Ubuntu 😂

Yes, we will be getting rid of the Node.js requirement in the coming weeks. When that happens, would we just need to update the it to `FROM debian-slim` or something?

contributing/Dockerfile
matthewmueller: Based on your link, it seems like we could update this to `--output -`? 

I have a slight preference towards curl for no other reason than I'm a lot more familiar with it.

matthewmueller:
I'm sold. Let me know what you think of the above, otherwise I think it's pretty much ready to go!

contributing/Dockerfile
wheinze: > Yes, we will be getting rid of the Node.js requirement soon. When that happens, would we just need to update the it to `FROM debian-slim` or something?

When we reach this state, I would recommend to switch over to [`golang:1.18.3`](https://hub.docker.com/_/golang), which is also based on Debian or preferable even `golang:1.18.3-alpine` which in the end saves us time as we don't need to setup Go anymore.

> An alpine image would be :chefkiss:

Would volunteer to work on that ;)

contributing/Dockerfile
wheinze: Hmm... thought I gave it a shot, but must have mixed up something back then. `curl` seems to be working fine with this solution, so let's just keep it.

contributing/Dockerfile
matthewmueller: Looks good! Sounds like there's a good migration path then.

> Would volunteer to work on that ;)
Appreciate it! You're quickly becoming my Devops guru 🧘‍♂️

Do you think there is an advantage to having multiple Dockerfiles in contributing? I'm tempted to just have one that's the most optimized we can make it.

matthewmueller:
Excellent, thank you Waldemar!

PR #142, opened 2022-06-19
replace redirect with case-insensitive routing

vito:
👋 Loving this project so far!

I'm switching an [app](https://github.com/vito/bass-loop/tree/bud) over to it to see how it goes and I noticed that Bud normalizes paths to lowercase. This is a problem for my app because it uses base64-encoded digests to route to content-addressed objects. I could switch to a case-insensitive database lookup, but that introduces the admittedly low possibility of collisions.

I've seen other websites in the wild that use case sensitive routes, for example url shorteners, so I don't think this is too unusual.

This PR removes the redirect and instead changes the router to use a case-insensitive match, since I figure the goal is for URLs to be case-insensitive. I'm not too opinionated here, just wanted to get a PR out to spark discussion. :)

Another option could be to convert casing for only the static portions of the path based on the route, but that's a bigger lift.

Thanks for making Bud!


matthewmueller:
Thrilled to hear Alex! It's so cool seeing Bass Loop using Bud. Bass also looks awesome. Looks like you're bringing the simplicity of functional programming to deploying infrastructure!

> This is a problem for my app because it uses base64-encoded digests to route to content-addressed objects. I could switch to a case-insensitive database lookup, but that introduces the admittedly low possibility of collisions.

Ah, I can totally see the problem. Something like: `TUlUIExpY2Vuc2UKCkNvcHl` gets redirected to `tuluiexpy2vuc2ukcknvchl`.

> This PR removes the redirect and instead changes the router to use a case-insensitive match, since I figure the goal is for URLs to be case-insensitive. I'm not too opinionated here, just wanted to get a PR out to spark discussion. :)

Thanks for opening this PR! I originally did this intentionally for SEO purposes, so `https://github.com/livebud/bud` and `https://github.com/livebud/BUD` wouldn't be treated as different routes in your analytics dashboards.

I had a look at what Github does and I think I like their approach:
1. They are case-insensitive, so `curl https://github.com/livebud/bud` and `curl https://github.com/livebud/BUD` render the same HTML.
2. Both HTML responses include `<link rel="canonical" href="https://github.com/livebud/bud" />` in their `<head>` to say that the lowercase version is the canonical route.

So I think your PR makes sense once the tests pass! 

If you feel like adding 2., you can adjust the default layout in [this file](https://github.com/livebud/bud/blob/2419421d5c3326f4eed9ed39e23e7ffa38657978/runtime/view/ssr/svelte.ts#L47). I'm not sure we're passing enough context to the views yet to provide a full URL like that, so this could be a follow-up task once I revisit views for layouts, error pages, etc.

vito:
Thanks. :)

> 2\. Both HTML responses include `<link rel="canonical" href="https://github.com/livebud/bud" />` in their `<head>` to say that the lowercase version is the canonical route.

Cool, that makes sense.

> So I think your PR makes sense once the tests pass!

Great! I got the tests passing, but I'm not sure of my approach. There were a few places that iterated over bytes, not runes. The "α" (`"\xce\xb1"`) route would be inserted fine, but then inserting "β" (`"\xce\xb2"`) would see`"\xce"` as a common prefix and set "α" and "β" as children of `path:"\xce"` in the routing tree. This somehow led to `/β` being returned for requests to `/α`.

So I changed `Tokens` to operate on runes rather than bytes, which affects `Size`, `At`, and `Split`. Took a while to figure out!

I'm interested in doing 2., but pushing just these changes for now; I'll try to find time to do it soon, so feel free to wait or merge as-is (if they actually pass in CI).

package/router/lex/tokens.go
```diff
@@ -41,9 +42,10 @@ func (tokens Tokens) At(i int) string {
 	for _, token := range tokens {
 		switch token.Type {
 		case PathToken, SlashToken:
-			for j := 0; j < len(token.Value); j++ {
+			chars := []rune(token.Value)
+			for j := 0; j < len(chars); j++ {
```
matthewmueller, line 46: Did you try ranging over `token.Value`? IIRC, you can do:

```go
for char := range token.Value {
  // do something
}
```

And it'll range over the runes.

matthewmueller: https://stackoverflow.com/a/18130921/145435

package/router/radix/tree.go
```diff
@@ -318,12 +318,10 @@ func matchExact(token lex.Token) matchFn {
 		if len(path) < rlen {
 			return -1, nil
 		}
-		for ; index < rlen; index++ {
-			if path[index] != route[index] {
-				return -1, nil
-			}
+		if !strings.EqualFold(path[:rlen], route) {
```
matthewmueller, line 321: TIL EqualFold!

package/router/radix/tree_test.go
```diff
@@ -273,6 +274,20 @@ func TestMatch(t *testing.T) {
...
 }
 
+func TestMatchUnicode(t *testing.T) {
+	ok(t, &test{
+		inserts: []*insert{
+			{route: "/α"},
+			{route: "/β"},
+		},
+		requests: []*request{
+			{path: "/α", route: "/α"},
```
matthewmueller, line 284: Could we add a request to a capital cased unicode character to see if it will route to the lowercase version?

For example, according to https://en.wikipedia.org/wiki/Greek_alphabet, the following should now work:

- /Δ => /δ

matthewmueller (changes requested):
A few minor tweaks then I think we're good to go!

package/router/lex/tokens.go
vito: Nice, I'll switch to that - I thought about it but didn't recognize that it "did the right thing" with the index key, i.e. skipping numbers for multi-byte runes.

matthewmueller (approved):
Nice work, thanks Alex!

PR #147, opened 2022-06-21
support HandlerFunc actions

vito:
As discussed in #143: if an action matches the `http.HandlerFunc` signature, use it directly. Route parameters are available in `r.URL.Query()`, and a test covers that as well.

Let me know if anything else is needed here!


matthewmueller:
This. is. awesome. Thanks Alex!!

Code looks great. Also cool to see how few changes you made to add this feature. 

Really nice work! 👏

PR #150, opened 2022-06-25
support glob embeds

vito:
I needed glob support for embedding migrations (`*.sql`) so here ya go! Feedback welcome as always. ✌️

### DIFF
--- a/internal/imhash/imhash_test.go
+++ b/internal/imhash/imhash_test.go
@@ -62,3 +62,54 @@ func TestAppHash(t *testing.T) {
 	is.Equal(len(hash2), 11)
 	is.True(hash1 != hash2)
 }
+
+func TestAppHashEmbed(t *testing.T) {
+	is := is.New(t)
+	ctx := context.Background()
+	dir := t.TempDir()
+	td := testdir.New(dir)
+	td.Files["controller/controller.go"] = `
+		package controller
+		import "embed"
+		//go:embed *.sql
+		var Migrations embed.FS
+		type Controller struct {}
+		func (c *Controller) Index() string {
+			return "Hello Users!!"
+		}
+	`
+	td.Files["controller/1.up.sql"] = `CREATE UNIVERSE;`
+	td.Files["controller/1.down.sql"] = `DROP UNIVERSE;`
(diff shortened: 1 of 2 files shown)


internal/imhash/imhash_test.go
```diff
@@ -62,3 +62,54 @@ func TestAppHash(t *testing.T) {
...
+
+func TestAppHashEmbed(t *testing.T) {
+	is := is.New(t)
+	ctx := context.Background()
+	dir := t.TempDir()
+	td := testdir.New(dir)
+	td.Files["controller/controller.go"] = `
+		package controller
+		import "embed"
+		//go:embed *.sql
```
matthewmueller, line 74: Would a list of embeds work as well? For example,

```go
//go:embed image/* template/*
//go:embed html/index.html
var content embed.FS
```

Source: https://pkg.go.dev/embed#hdr-Directives

If not, maybe we could add that to this PR or leave a more specific TODO to support a list of embeds. What do you think?

matthewmueller (commented):
Oh nice! This looks great Alex. Left one comment.

internal/imhash/imhash_test.go
vito: Looks like it works as-is; I've expanded the test to cover more cases.

internal/imhash/imhash_test.go
```diff
@@ -62,3 +62,106 @@ func TestAppHash(t *testing.T) {
...
+		var Migrations embed.FS
+		type Controller struct {}
+		func (c *Controller) Index() string {
+			return "Hello Users!!"
+		}
+	`
+	td.Files["controller/1.up.sql"] = `CREATE UNIVERSE;`
+	td.Files["controller/1.down.sql"] = `DROP UNIVERSE;`
+	td.Files["controller/html/index.html"] = `<blink>hi</blink>`
+	td.Files["controller/css/a.css"] = `body { font-family: Comic Sans; }`
```
matthewmueller, line 85: 🤡

matthewmueller:
Awesome, thanks for checking that. This looks good to go!

PR #167, opened 2022-06-28
internal/cli/create: fix scaffolding when outside of $GOPATH.

matthewmueller:
Closes: #166

PR #176, opened 2022-06-30
internal/cli: write to dir instead of temp

matthewmueller:
I originally wanted to write to tempdir because I didn't want failures to result in half built directories. Since folks are [running into problems](https://github.com/livebud/bud/pull/167#issuecomment-1168279700), I think its worth revisiting this. 

This PR will now just write into the intended directory.


matthewmueller:
@012e, can you try `create` again on main? I want to see if this addresses your issue.

Issue #45, opened 2022-05-15
Provide a nicer error message for unsupported Go versions


one-aalam:
After creating a fresh project, the command `bud run` fails with import related errors 
<img width="851" alt="bud-h" src="https://user-images.githubusercontent.com/532906/168464730-c89625c4-95c2-4fb6-858e-c6df6f3b48aa.png">

Looks like the scaffolded structure is missing some important directories/files, but I'm not sure why that's happening
<img width="846" alt="bud-h" src="https://user-images.githubusercontent.com/532906/168464840-3f820fd0-84c3-4263-8aaa-d79ccd95e6c0.png">

matthewmueller:
Oh, interesting! I guess you're running a version of Go that's less than 1.16? Bud relies heavily on `io/fs` and will take advantage of generics in the future, so I suggest running Go 1.18+ if you can.

That's an awful error nonetheless, I'll improve that. Let me know what version you're running with `go version`.

one-aalam:
Hey @matthewmueller Thank you. Bud's running flawlessly with the `golang` upgrade.

joshuatalb:
I was also caught out by this. I've made a small PR (#53) to add a comment in the `README.md` about this.

jmaister:
It could help that the `cli` checks for the appropriate version of go and exit with a nice message.

Maybe when running on Windows, I also read in other issues that there are problems with it too.

matthewmueller:
Agreed. Accepting PRs!

PR #193, opened 2022-07-04
Feature/check golang version

jfmario:
I'm not sure if this is the best way of doing this or not.

This is related to issue #45 .

I added a small package with a representation of the version returned by `runtime.Version()`, including a way to compare versions. This is checked at the beginning of handling CLI input.

If the running version is less than the hard-coded required version, then a message is printed and the process stops. If the version returned was a commit hash or anything other than a normal version of go, then the process continues after a warning message.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -3,10 +3,13 @@ package cli
 import (
 	"context"
 	"errors"
+	"fmt"
+	"os"
 
 	"github.com/livebud/bud/internal/cli/bud"
 	"github.com/livebud/bud/internal/cli/build"
 	"github.com/livebud/bud/internal/cli/create"
+	"github.com/livebud/bud/internal/cli/goversion"
 	"github.com/livebud/bud/internal/cli/newcontroller"
 	"github.com/livebud/bud/internal/cli/run"
 	"github.com/livebud/bud/internal/cli/toolbs"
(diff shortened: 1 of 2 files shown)


jfmario:
Hmm, this ran locally on my mac and on my EC2 (amazon linux). I will try to look into what I might have done to cause all these failures.

matthewmueller:
Hey @jfmario, thanks so much for your contribution to Bud! You can fix the CI error by running `make go.fmt` to format your Go code.

Hey so, I started a review but then realized there were quite a few things to adjust and I thought it would be easier to open another PR instead. I did so in #194. I really hope you don't mind!

The main things that were changed:
- Already using the x/mod/semver package, so let's just re-use that code (try `go list -f '{{ .Imports }}' golang.org/x/mod/module`
- From reading about what other types of `runtime.Version()` values there could be, I realize that it's mostly dev builds which I think will mostly be higher than Go 1.17. I think it's worth letting them pass and seeing what happens.

I hope this doesn't discourage you from opening other PRs. Thank you for your interest in contributing!

jfmario:
> I hope this doesn't discourage you from opening other PRs. Thank you for your interest in contributing!

Not discouraged! Thank you for taking the time to look at this and explaining your thinking.

I'm still somewhat new in Go, and I haven't contributed to open source projects so I expect PRs to get heavily edited or passed over for other approaches! Because I like Go so much and have benefitted from Django and Laravel at other times, I really like the idea of Bud. I'll continue digging into the framework, learning about the go ecosystem, and will look for opportunities to contribute where I can.

Issue #55, opened 2022-05-16
Add bud new controller <resource> {create,new,edit,update,delete} actions


TechplexEngineer:
Thanks for making bud!

I must be doing something wrong here. I'd like to scaffold create, read, update and delete actions. 

It seems like only index and show are implemented, is that correct? (when running `bud new controller ...`

matthewmueller:
@TechplexEngineer thanks for checking out Bud. You're not doing something wrong!

This is slightly embarrassing, but scaffolding was the last thing I wrote and I was so over keeping bud a secret at that point, that I implemented the bare minimum to record the demo.

Sincere apology for the confusion, I meant to fix this up before anyone noticed.

matthewmueller:
The PR has been started in #75

PR #195, opened 2022-07-07
Added controller action generator for create, new, edit, update, delete

dangerzone:
Addresses issue https://github.com/livebud/bud/issues/55 since https://github.com/livebud/bud/pull/75 has become stale.

In addition to the changes made by @fergloragain, this PR fixes some errors present in that PR and adds views for edit and new actions. It also makes the action creation error more helpful.

The views aren't terribly pretty though since I'm not one with the CSS.

### DIFF
--- a/internal/cli/newcontroller/newcontroller.go
+++ b/internal/cli/newcontroller/newcontroller.go
@@ -38,9 +38,17 @@ var indexView string
 //go:embed view_show.gotext
 var showView string
 
+//go:embed view_edit.gotext
+var editView string
+
+//go:embed view_new.gotext
+var newView string
+
 var views = map[string]string{
 	"index": indexView,
 	"show":  showView,
+	"edit":  editView,
+	"new":   newView,
 }
 
 type State struct {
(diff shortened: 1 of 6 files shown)


internal/cli/newcontroller/newcontroller.go
```diff
@@ -170,13 +219,15 @@ func (c *Command) views(controller *Controller) (views []*View) {
 }
 
 func (c *Command) view(controller *Controller, action *Action) *View {
+	fmt.Println(action.Name)
```
matthewmueller, line 222: Remove fmt

matthewmueller (changes requested):
Thanks so much for for working on #75 @dangerzone!

I pulled it down and tried:

```
bud new controller posts new show index create delete update edit
```

But ran into a concurrency error. The problem is the vfs.Memory in the scaffolder isn't concurrency-safe. Feel free to take a stab at fixing this, otherwise I'll get that fixed over the weekend.

internal/cli/newcontroller/newcontroller.go
dangerzone: If we wrap it in a mutex, that'd defeat the purpose of having it be in a goroutine. For now, I say we run the scaffolding sequentially instead of concurrently. This isn't exactly a performance critical hot-path anyway and there are lower hanging fruit for us to occupy our time with ;)

dangerzone: I removed the concurrent calls and it works as expected

Issue #138, opened 2022-06-15
Renaming or deleting views should trigger a recompile


matthewmueller:
Given the following file directory:

```
view/
  show.svelte
```

If you try renaming `show.svelte` to `_show.svelte` while running `bud run`, the change doesn't register.

This is because bud currently treats all view changes as incrementally reloadable, but in this case, we need to regenerate the controller.

Renames and deletes of views should trigger a full-page reload.

PR #197, opened 2022-07-08
Renaming or deleting views should trigger a recompile

dangerzone:
Refactored watcher to include type of change event. Non-edit changes to view files triggers an app restart.

Addresses issue https://github.com/livebud/bud/issues/138


matthewmueller:
Thanks @dangerzone for your help here! While your PR totally would have worked, I wanted to pass the entire event through. We also need tests for changes introduced :) 

Closing in favor of #212

PR #203, opened 2022-07-14
Scaffold remaining actions and test protocol

### LINKED ISSUE #55: Add bud new controller <resource> {create,new,edit,update,delete} actions

matthewmueller:
This PR adds support for the remaining actions. For example, `bud new controller posts create update delete edit new`.

It also adds method overriding to support PATCH and DELETE requests from form elements: https://guides.rubyonrails.org/form_helpers.html#how-do-forms-with-patch-put-or-delete-methods-work-questionmark

Closes #195 
Closes #55 
Closes #75


jfmario:
I am taking a look at this and some of the test errors. 

One of them expects an empty object when the sample controller returns `{"id":10}`. I think this is expected and I can resolve by editing the expectation in `newcontroller_test.go`.

Another error is because of the 404 returned to `POST /posts/10?_method=patch`

jfmario:
I was poking around at this, and looking at some of the test failures.

One takes place on `GET /posts/10` because the test expects an empty object but the controller returns `{"id":10}` which I think is expected.

The other related error is that `PATCH /posts/10` works but not `POST /posts/10?_method=patch`. 

I took a stab at this error in the link below, but I'm sure its not in the correct location.

https://github.com/jfmario/bud/blob/scaffold-actions/package/middleware/method_overload.go

Is this the general idea, though, to put some middleware in front of the router that alters the request method? It was working when I tested locally with `?_method=PATCH`.

matthewmueller:
Awesome, thanks for looking into this @jfmario. I'm going to try and wrap this up today. 

> Is this the general idea, though, to put some middleware in front of the router that alters the request method? It was working when I tested locally with ?_method=PATCH

Yep, exactly. More info here: https://guides.rubyonrails.org/form_helpers.html#how-do-forms-with-patch-put-or-delete-methods-work-questionmark

I'll pull in the code you wrote and make some minor adjustments, but you've got the right idea!

Issue #149, opened 2022-06-23
Can't create routes named `index`, `show`


012e:
# Reproduce

1. Create a new project `bud create project`
2. Create a root route  **(important)** `bud new controller anything:/ index show` 
3. `bud new controller index` or `bud new controller show` (with or without views)
4. `bud run` will give an error:
```
| conjure: generate "bud/.app/main.go". conjure: generate "bud/.app/program/program.go". program: unable to wire. di: unable to wire "bud-app/bud/program".loadApp function. di: unable to find definition for param "bud-app/bud/.app/controller".*Controller in "bud-app/bud/.app/web".*Server . parser: unable to find declaration for "bud-app/bud/.app/controller".Controller in "bud/.app/web/web.go". bud/.app/controller/controller.go:15:2: Index redeclared in this block
	previous declaration at bud/.app/controller/controller.go:13:2
```

matthewmueller:
Thanks for the report @012e! That's strange, but I'm not too surprised since it's untested. Just want to double-check. The `1/` directory isn't related to #144 right?

012e:
Yes, `1` is just the module name.

012e:
My previous report was wrong so I rewrote it

matthewmueller:
Thanks! I can reproduce this one. Are you blocked by this one @012e? I may bundle this one together with reworking the controller generator to be middleware.

012e:
> Are you blocked by this one @012e?

Not really, I don't need this

dangerzone:
I was not able to reproduce this with the latest bud version 0.2.0

On a new bud project:

`bud new controller foobar:/ index show
bud new controller index
`

Worked as expected with no errors

jfmario:
It is trying to create this during file generation (at the top of `controller/controller.go`).

```go
// Controller struct
type Controller struct {
	Index *IndexAction
	Show *ShowAction
	Index *IndexController
}
```

I'm looking at the `.gotext` files and experimenting with how feasible it would be to do the following instead:

```go
// Controller struct
type Controller struct {
	IndexAction *IndexAction
	ShowAction *ShowAction
	IndexController *IndexController
}
```

PR #214, opened 2022-07-16
Issue/149

jfmario:
This addresses #149.

Appends the suffix "action" to all actions and the suffix "controller" to all nested controllers.

I'm sure there's downsides to this approach, so I'm interested in your thoughts. I was able to verify that this solves the problem listed in the issue, both at the root and at other nested levels.

Tests fail, as assertions expect certain names to exist. I can look at that if you agree with this suffix approach.

### DIFF
--- a/framework/controller/controller_test.go
+++ b/framework/controller/controller_test.go
@@ -2709,33 +2709,43 @@ func TestProtocol(t *testing.T) {
 }
 
 // Asserts that a root index action is compatible with an index
-// controller.
+// controller. Also asserts that a posts index action is compatible
+// with a posts/index controller.
 func TestIndexControllerWithRootIndexAction(t *testing.T) {
 	is := is.New(t)
 	ctx := context.Background()
 	dir := t.TempDir()
 	td := testdir.New(dir)
-	// enable svelte
-	td.NodeModules["svelte"] = versions.Svelte
 	// root controller with index action
 	td.Files["controller/controller.go"] = `
 		package controller
 		type Controller struct{}
-		func (c *Controller) Index() bool {
-			return true
+		func (c *Controller) Index() []string {
+			return []string{}
 		}
-	`
-	// root index view
-	td.Files["view/index.svelte"] = `
-		<h1>Home Page</h1>
 	`
 	// index controller with index action
 	td.Files["controller/index/controller.go"] = `
 		package index
-		type Widget struct{}
 		type Controller struct{}
-		func (c *Controller) Index() ([]*Widget, error) {
-			return []*Widget{}, nil
+		func (c *Controller) Index() []string {
+			return []string{}
+		}
+	`
+	// posts controller with index action
+	td.Files["controller/posts/controller.go"] = `
+		package posts
+		type Controller struct{}
+		func (c *Controller) Index() []string {
+			return []string{}
+		}
+	`
+	// posts/index controller with index action
+	td.Files["controller/posts/index/controller.go"] = `
+		package index
+		type Controller struct{}
+		func (c *Controller) Index() []string {
+			return []string{}
 		}
 	`
 	is.NoErr(td.Write(ctx))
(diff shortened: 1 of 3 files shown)


matthewmueller:
Thanks for the PR! The main consideration here was the thought that this may be a public API when we expose custom routing. 

So you could you have a routes.go file that has something like:

```go
func Routes(router *router.Router, cmap *controller.Map) {
  router.Get("/", cmap.Users.Index)
}
```

I haven't put too much thought into how to address #149 and still have a nice public API, but given this new information, would you change anything about this PR?

jfmario:
If this is going to be a public API then the PR isn't acceptable. 

I'm not sure how to get around #149, but I'll try to think about it more tomorrow. It's certainly OK to have some things just now be allowed. For example a request to "/" and a request to "/index.html" are usually the same, so I think it's fine to document that controller packages called things like "index" and "show" aren't supported at this time (this is at ANY level of nesting, not just root).

If it's actually important to have a `/index` route go to a controller, then having a custom `routes.go` would allow that and maybe that's the best answer for #149.

jfmario:
Another thing i haven't tried is `cmap.UsersController.Index` (only suffixing the controller). It's not as nice as `cmap.Users.Index`, but it's not ugly either in my view

matthewmueller:
> Another thing i haven't tried is cmap.UsersController.Index (only suffixing the controller). It's not as nice as cmap.Users.Index, but it's not ugly either in my view

I think this might be the good solution! Want to update the PR + tests? One thing to test would be how the root controller (e.g. controller/controller.go) would behave with an index controller (e.g. controller/index/controller.go). I think it should work, but it'd be nice to get a test for it.

jfmario:
Yes, I'll work with that and get back to you!

jfmario:
@matthewmueller I made the changes, and added a test case that passes (at least locally).

framework/controller/controller_test.go
```diff
@@ -2707,3 +2707,54 @@ func TestProtocol(t *testing.T) {
...
+		HTTP/1.1 200 OK
+		Transfer-Encoding: chunked
+		Content-Type: text/html
+	`))
+	res, err = app.GetJSON("/index")
+	is.NoErr(err)
+	is.NoErr(res.DiffHeaders(`
+		HTTP/1.1 200 OK
+		Content-Type: application/json
+	`))
```
matthewmueller, line 2758: Let's use `res.Diff` here since we're making a JSON request and can compare the whole response.

framework/controller/controller_test.go
```diff
@@ -2707,3 +2707,54 @@ func TestProtocol(t *testing.T) {
...
+		func (c *Controller) Index() ([]*Widget, error) {
+			return []*Widget{}, nil
+		}
+	`
+	is.NoErr(td.Write(ctx))
+	cli := testcli.New(dir)
+	app, err := cli.Start(ctx, "run")
+	is.NoErr(err)
+	defer app.Close()
+	res, err := app.Get("/")
```
matthewmueller, line 2746: Switch to `GetJSON("/")` since we don't really need to test HTML responses if JSON responses work.

framework/controller/controller_test.go
```diff
@@ -2707,3 +2707,54 @@ func TestProtocol(t *testing.T) {
...
+	is.NoErr(err)
+	is.NoErr(res.DiffHeaders(`
+		HTTP/1.1 200 OK
+		Transfer-Encoding: chunked
+		Content-Type: text/html
+	`))
```
matthewmueller, line 2752: use `res.Diff(...)` to test the whole response

framework/controller/controller_test.go
```diff
@@ -2707,3 +2707,54 @@ func TestProtocol(t *testing.T) {
...
+		<h1>Home Page</h1>
+	`
+	// index controller with index action
+	td.Files["controller/index/controller.go"] = `
+		package index
+		type Widget struct{}
+		type Controller struct{}
+		func (c *Controller) Index() ([]*Widget, error) {
+			return []*Widget{}, nil
+		}
```
matthewmueller, line 2739: Consider simplifying to the route you have above. Saves a line or two. We just need to test that routing works and there's no errors.

framework/controller/controller_test.go
```diff
@@ -2707,3 +2707,54 @@ func TestProtocol(t *testing.T) {
...
+	// root controller with index action
+	td.Files["controller/controller.go"] = `
+		package controller
+		type Controller struct{}
+		func (c *Controller) Index() bool {
+			return true
+		}
+	`
+	// root index view
+	td.Files["view/index.svelte"] = `
```
matthewmueller, line 2729: Can remove if you're testing JSON

framework/controller/controller_test.go
```diff
@@ -2707,3 +2707,54 @@ func TestProtocol(t *testing.T) {
...
+
+// Asserts that a root index action is compatible with an index
+// controller.
+func TestIndexControllerWithRootIndexAction(t *testing.T) {
+	is := is.New(t)
+	ctx := context.Background()
+	dir := t.TempDir()
+	td := testdir.New(dir)
+	// enable svelte
+	td.NodeModules["svelte"] = versions.Svelte
```
matthewmueller, line 2719: Can remove if there's no view.

framework/web/loader.go
```diff
@@ -168,7 +168,16 @@ func (l *loader) loadActionRoute(basePath, actionName string) string {
...
-	baseCaller := text.Dot(text.Title(basePath))
+
+	splitPath := strings.Split(text.Title(basePath), " ")
+	for i, controller := range splitPath {
+		// append controller to every controller base name
+		if len(controller) > 0 {
+			splitPath[i] = controller + "Controller"
+		}
+	}
+	baseCaller := strings.Join(splitPath, ".")
```
matthewmueller, line 179: Hmm... this isn't quite the same behavior as what's on the left. IIRC:

```
text.Title("users/posts/comments") => UsersPostsComments
text.Dot("UsersPostsComments") => Users.Posts.Comments
```

I _think_ there's no cases where `splitPath` actually has more than one item:

```
strings.Split(text.Title("users/posts/comments"), " ") => []string{"UsersPostsComments"}
```

I'm kind of surprised the tests would be passing though. Maybe I'm missing something. Can you double-check this and maybe write a test above with a deeply nested controller?

matthewmueller (commented):
Thanks for working on this @jfmario!

PR #225, opened 2022-07-22
framework/view: experiment to extract out svelte into a swappable "viewer"

matthewmueller:
This PR is an experiment to support swapping out the view renderers. Currently we have just one Svelte "viewer", but it'd be nice to be able to define your own viewers. This would allow for other viewers like GoHTML, HTML and React.

I haven't decided if this interface should be exposed as a generator yet, but at the very least the internal interfaces will be capable of supporting other types of renderers. If it does make sense to expose it, what should we call this? Viewers? Renderers? Templates? Something else?


framework/view/loader.go
```diff
@@ -87,6 +86,7 @@ func (l *loader) Load(ctx context.Context) (state *State, err error) {
 		l.imports.AddNamed("budclient", "github.com/livebud/bud/package/budclient")
 	}
 	l.imports.AddNamed("viewrt", "github.com/livebud/bud/framework/view/viewrt")
+	l.imports.AddNamed("svelte", l.module.Import("viewer/svelte"))
```
matthewmueller, line 89: WIP: hardcoded

framework/view/viewrt/viewer.go
```diff
@@ -0,0 +1,14 @@
...
+
+	"github.com/livebud/bud/package/router"
+)
+
+type Viewer interface {
+	// TODO: remove handler
+	Handler(route string, props interface{}) http.Handler
+	Render(w http.ResponseWriter, viewPath string, props interface{})
+	Serve(router *router.Router)
+}
```
matthewmueller, line 14: This interface still needs some work, but I just wanted to get something working.

internal/virtual/file.go
```diff
@@ -0,0 +1,67 @@
+package virtual
+
+import (
```
matthewmueller, line 3: This is a copy from both conjure and fscache. I'd like to consolidate those implementations into virtual.

package/budclient/client.go
```diff
@@ -5,15 +5,20 @@ import (
...
+	"github.com/livebud/bud/internal/virtual"
+	"github.com/livebud/bud/package/js"
 	"github.com/livebud/bud/package/socket"
 )
 
 type Client interface {
+	js.VM
+	fs.FS
 	Render(route string, props interface{}) (*ssr.Response, error)
 	Proxy(w http.ResponseWriter, r *http.Request)
```
matthewmueller, line 23: This PR would make Render and Proxy obsolete.

package/di/function.go
```diff
@@ -249,7 +249,9 @@ func maybePrefixParam(param *Type, input *Variable) string {
 	// input is not an interface.
 	// TODO: This is hacky because input.Kind can't tell if it's a pointer to a
 	// struct vs. a struct, so we need extra logic to tell.
-	if !sameImport && isInterface(param.kind) && !isInterface(input.Kind) && !strings.HasPrefix(input.Type, "*") {
+	// TODO: input.Kind.String() != "unknown" is flaky. The unknown kind should
+	// be removed altogether.
+	if !sameImport && isInterface(param.kind) && !isInterface(input.Kind) && input.Kind.String() != "unknown" && !strings.HasPrefix(input.Type, "*") {
```
matthewmueller, line 254: Fixed a bug in DI where aliasing an interface with an external interface would cause compilation errors. The fix added to a hack that should be cleaned up eventually.

PR #236, opened 2022-08-03
Initial custom generator support

matthewmueller:
This PR contains a prototype for the custom generator support. Supporting custom generators earlier will speed up prototyping features like transforms, viewers, tailwind support as well as open up the ecosystem.

The current design looks like this:

```
$APP
  generator/
    tailwind/
      tailwind.go
    transform/
      markdoc/
        markdoc.go
```

Where `$APP/generator/tailwind/tailwind.go` looks something like this:

```go
package tailwind

type Generator struct {
  // dependencies....
}

// Register generators scoped to the directory
func (g *Generator) Register(genfs *overlay.FileSystem, flag *framework.Flag) {
        // Generate `tailwind.css` when requesting `bud/internal/generator/tailwind/tailwind.css`
	genfs.FileGenerator("tailwind.css", GenerateFile(ctx context.Context, fsys overlay.F, file *overlay.File) error {
		fmt.Println("running generator!")
		file.Data = []byte("/** tailwind **/")
		return nil
	})
        // ... Generate more files or directories ...
	return nil
}
```

To get this working, this PR uses a plugin system over RPC under the hood, similar to https://github.com/hashicorp/go-plugin.

**Todo**

- [x] Hook up Register to the `generator.gotext`
- [x] Figure out how to reduce the number of expensive calls to GenerateDir. We should probably call this function at most once within conjure.
- [x] Figure out how to properly cleanup after spawning the custom generator plugin.
- [x] Finish up `generator.gotext` adding dependency injection support to the generators
- [x] Write tests


barelyhuman:
Might wanna consider file locking options (in case someone wishes to write multiple generators to write/append to the same file) (totally overthinking but will be a good option)

> plugin system works by launching subprocesses and communicating over RPC

You did mention cleanup but then if they are running in parallel (go routines) don't you think this would be heavy on the memory? (just assumption, might need correction)

matthewmueller:
@barelyhuman thanks for your thoughts! 

> Might wanna consider file locking options (in case someone wishes to write multiple generators to write/append to the same file) (totally overthinking but will be a good option)

Right now the generators run serially and all in-memory before flushing to disk. This will be important once I parallelize filepath walking, but I'll use mutexes instead of file locks.

> You did mention cleanup but then if they are running in parallel (go routines) don't you think this would be heavy on the memory? (just assumption, might need correction)

It would only be one goroutine, so I don't think this will be a big deal.

barelyhuman:
Make sense, 

> I'll use mutexes instead of file locks

Cool! 

> Maybe Go plugins once they're faster to compile and have less issues

I was looking at the other RPC options for plugins as well and considering how VM's work, most of the time if you are adding a VM related file processing, you can't really parallelise without having one VM instance per routine so communication between them is going to get messy with quite a few channels. 

Though I think that's a problem for later since the generators are serial right now

Amazing work! Let me know if I could help somewhere :)

Issue #240, opened 2022-08-05
Improve onboarding when outside $GOPATH


matthewmueller:
Currently leaning towards:

- If in `$GOPATH`, no change in behavior. Infer the module name.
- If outside of of `$GOPATH`, follow what Wails does, unless...
  -  you've specified a `--module=<name>`

Context: https://www.reddit.com/r/golang/comments/wf95gr/comment/iitsjrd/?utm_source=reddit&utm_medium=web2x&context=3

matthewmueller:
Closed by #242!

PR #242, opened 2022-08-08
Auto infer module when `bud create` outside $GOPATH

012e:
Implements #240

### DIFF
--- a/internal/cli/create/create.go
+++ b/internal/cli/create/create.go
@@ -110,9 +110,9 @@ func (c *Command) loadModule() *Module {
 	if module.Name == "" {
 		// Try inferring the module name from the directory
 		module.Name = mod.Infer(c.absDir)
-		// if can't infer then default it to `changeme`
+		// if can't infer then default it to `change.me`
 		if module.Name == "" {
-			module.Name = "changeme"
+			module.Name = "change.me"
 		}
 	}
 	// Autoquote the module name since
--- a/internal/cli/create/create.go
+++ b/internal/cli/create/create.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/livebud/bud/internal/bail"
 	"github.com/livebud/bud/internal/cli/bud"
-	"github.com/livebud/bud/internal/format"
 	"github.com/livebud/bud/internal/scaffold"
 	"github.com/livebud/bud/internal/versions"
 	mod "github.com/livebud/bud/package/gomod"
@@ -111,14 +110,9 @@ func (c *Command) loadModule() *Module {
 	if module.Name == "" {
 		// Try inferring the module name from the directory
 		module.Name = mod.Infer(c.absDir)
+		// if can't infer then default it to `changeme`
 		if module.Name == "" {
-			// Fail that you need to pass in a module path
-			c.bail.Bail(format.Errorf(`
-			Unable to infer a module name. Try again using the module <path> name.
-
-			For example,
-				bud create --module=github.com/my/app %s
-		`, c.Dir))
+			module.Name = "changeme"
 		}
 	}
 	// Autoquote the module name since
--- a/internal/cli/create/create_test.go
+++ b/internal/cli/create/create_test.go
@@ -1,7 +1,9 @@
 package create_test
 
 import (
+	"bufio"
 	"context"
+	"os"
 	"testing"
 
 	"github.com/livebud/bud/internal/cli/testcli"
@@ -10,7 +12,15 @@ import (
 	"golang.org/x/mod/modfile"
 )
 
-func TestCreateOutsideGoPathError(t *testing.T) {
+func fileFirstLine(filePath string) string {
+	file, _ := os.Open(filePath)
+	defer file.Close()
+	scanner := bufio.NewReader(file)
+	line, _ := scanner.ReadString('\n')
+	return line
+}
+
+func TestCreateOutsideGoPath(t *testing.T) {
 	is := is.New(t)
 	ctx := context.Background()
 	dir := t.TempDir()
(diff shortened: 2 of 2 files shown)


012e:
@matthewmueller should I rename my branch?
https://github.com/livebud/bud/blob/e23cd48a1a89ebaba2fbe8caf2d2cc5b9369ac45/Makefile#L253

internal/cli/create/create.go
```diff
@@ -111,14 +110,9 @@ func (c *Command) loadModule() *Module {
...
+		// if can't infer then default it to `changeme`
 		if module.Name == "" {
-			// Fail that you need to pass in a module path
-			c.bail.Bail(format.Errorf(`
-			Unable to infer a module name. Try again using the module <path> name.
-
-			For example,
-				bud create --module=github.com/my/app %s
-		`, c.Dir))
+			module.Name = "changeme"
```
matthewmueller, line 115: I think we should make it `change.me` because `changeme` isn't a valid module name. This doesn't appear to have any ramifications right now, but I don't see a big deal with generating a valid name. See below for an example:

```go
package main

import (
	"fmt"
	"os"

	"golang.org/x/mod/module"
)

func main() {
	if len(os.Args) <= 1 {
		fmt.Println("no args")
		return
	}
	modulePath := os.Args[1]
	err := module.CheckPath(modulePath)
	if err != nil {
		fmt.Printf("invalid path %s: %s\n", modulePath, err)
		os.Exit(1)
	}
	fmt.Printf("valid path %s\n", modulePath)
}
```

internal/cli/create/create_test.go
```diff
@@ -20,10 +30,13 @@ func TestCreateOutsideGoPathError(t *testing.T) {
 	cli := testcli.New(dir)
 	is.NoErr(td.NotExists(".gitignore"))
 	result, err := cli.Run(ctx, "create", dir)
-	is.In(err.Error(), `Try again using the module <path> name.`)
 	is.Equal(result.Stdout(), "")
 	is.Equal(result.Stderr(), "")
-	is.NoErr(td.NotExists(".gitignore"))
+	is.Equal(fileFirstLine(dir+"/go.mod"), "module changeme\n")
```
matthewmueller, line 35: filepath.Join(dir, "go.mod")

internal/cli/create/create_test.go
```diff
@@ -37,6 +50,7 @@ func TestCreateOutsideGoPathModulePath(t *testing.T) {
 	is.NoErr(err)
 	is.Equal(result.Stdout(), "")
 	is.Equal(result.Stderr(), "")
+	is.Equal(fileFirstLine(dir+"/go.mod"), "module github.com/my/app\n")
```
matthewmueller, line 53: same as above

matthewmueller:
Thanks for working on this @012e! Just reviewed.

> @matthewmueller should I rename my branch?

This looks like a bug in the recent Makefile E2E test not accounting for branches outside of `livebud/bud`.

I think the problem is that `auto-infer-module` branch doesn't exist on `livebud/bud`. It only exists on your branch.

Could you try changing the Makefile? It seems like we could use the `GITHUB_REPOSITORY` environment variable in Github Actions: https://docs.github.com/en/actions/learn-github-actions/environment-variables

So maybe something like the following:

```diff
BRANCH_NAME ?= $(or $(GITHUB_HEAD_REF),main)
+ GITHUB_REPO ?= $(or $(GITHUB_REPOSITORY),livebud/bud)
GOPATH := $(shell go env GOPATH)


e2e: e2e.bud.build


e2e.bud.build:
-	GOPRIVATE=github.com/livebud/bud go install github.com/livebud/bud@$(BRANCH_NAME)
+	GOPRIVATE=github.com/$(GITHUB_REPO) go install github.com/$(GITHUB_REPO)@$(BRANCH_NAME)
	git clone https://github.com/livebud/welcome
	( cd welcome && \
		npm install && \
		go mod tidy && \
-		GOPRIVATE=github.com/livebud/bud go get github.com/livebud/bud@$(BRANCH_NAME) \
+               GOPRIVATE=github.com/$(GITHUB_REPO) go get github.com/$(GITHUB_REPO)@$(BRANCH_NAME) \
	)
	$(GOPATH)/bin/bud -C welcome build
	./welcome/bud/app -h
```

012e:
> ```diff
> BRANCH_NAME ?= $(or $(GITHUB_HEAD_REF),main)
> + GITHUB_REPO ?= $(or $(GITHUB_REPOSITORY),livebud/bud)
> GOPATH := $(shell go env GOPATH)
> 
> 
> e2e: e2e.bud.build
> 
> 
> e2e.bud.build:
> -	GOPRIVATE=github.com/livebud/bud go install github.com/livebud/bud@$(BRANCH_NAME)
> +	GOPRIVATE=github.com/$(GITHUB_REPO) go install github.com/$(GITHUB_REPO)@$(BRANCH_NAME)
> 	git clone https://github.com/livebud/welcome
> 	( cd welcome && \
> 		npm install && \
> 		go mod tidy && \
> -		GOPRIVATE=github.com/livebud/bud go get github.com/livebud/bud@$(BRANCH_NAME) \
> +               GOPRIVATE=github.com/$(GITHUB_REPO) go get github.com/$(GITHUB_REPO)@$(BRANCH_NAME) \
> 	)
> 	$(GOPATH)/bin/bud -C welcome build
> 	./welcome/bud/app -h
> ```


After a while researching, I [found](https://github.com/golang/go/issues/50278) we can't `go install` a fork. Should we just remove the e2e test for now and create a new issue?

matthewmueller:
Thanks @012e for the research! I just updated the e2e target to hopefully address this.

matthewmueller:
Cheers, thanks Huy!

PR #244, opened 2022-08-09
budfs support

matthewmueller:
This is an experimental v3 of the generator filesystem. The goals are:
- Bake in dependency caching so running spawning subprocesses only occurs when dependent files change
- Fix minor issues around dir entries on generated files having a size of 0 or 0 for their file mode
- Prep for a custom generator API that depends on budfs
<s>- Join conjure and overlay into one bud-specific filesystem</s>
- Be able to serve files using the same filesystem without caching
- Better errors (conjure => budfs)


matthewmueller:
Going to merge this now and integrate in follow-up PRs. I tried integrating all at once, but that turned out to be way too deep of a rabbit hole.

The main changes this PR makes:
- Adds budfs to eventually replace overlay
- Adds genfs to eventually replace conjure
- Adds mergefs to eventually replace merged
- Adds cachefs to eventually replace fscache
- Adds `remotefs.Command` to spawn a remotefs server
- Add `treefs` underneath to get `fstest.TestFS` fully passing on the virtual filesystem implementation

These all build on top of the newer `virtual` package.

The goals from above remain the same.

Issue #253, opened 2022-08-20
Default CSS in each page


ralfiannor:
Instead applied a default css in each html page, why not to add the css style in the `example project` or `welcome page generator` only.

matthewmueller:
Hey @ralfiannor, thanks for trying Bud! 

The `default.css` implementation was a bit rushed, but I think the goal is still worthwhile:

**Original Goal:** When there's no other CSS on the page, Bud provides reasonable defaults. This stems from the fact that _every website_ adds `normalize.css`, `reset.css`, `preflight.css`, etc. 

A couple solutions I can think of right now.

**1. You can override `default.css` by creating a `public/default.css`** 

This was the original plan, by creating a `public/default.css`, it becomes the `default.css`. The way you'd reset is by creating an empty file. This may not work so well with CSS frameworks like Tailwind though.

**2. No default styles, instead you're expected to scaffold your CSS solution**

Something like:

```sh
bud new css tailwind
bud create --css=tailwind
```

**3. A default CSS stylesheet is added when you create a new project. You can delete it if you don't want it**

This way you get the default, but you can go back to no styling or replace it by deleting the generated file. You could additionally add scaffolders for Tailwind like in **2.**, but the difference is that **3.** starts with a default, where **2.** does not.

---

I'm leaning towards **3.** right now because I think it provides the most flexibility.

matthewmueller:
I remember @vito also being surprised about the `default.css`, do you have any thoughts?

vito:
I like 3 as well. I was a little surprised to have default CSS but I think it's a reasonable thing for Bud to do, since as you said just about every project starts with one, and I think it's good for Bud to cut out as much boilerplate as possible.

What I like about 3 is I can clearly see that it generated a default CSS for me. Thinking back what was more surprising was that it was happening without any indication, it was just sort of injected into the page for me.

One thing to consider is what happens when I bump to a new version of Bud that comes with a new default CSS. I guess I would expect the `default.css` to stay as-is since it's only at project generation time. I think that's preferable as long as there's some obvious way of upgrading it if the user wants, which might just mean copy/pasting from upstream.

matthewmueller:
> One thing to consider is what happens when I bump to a new version of Bud that comes with a new default CSS.

Good point! I also agree that a manual upgrade would be preferable to a potential styling surprise. Copy and paste for now would be fine since I don't see normalize.css changing much. 

Let's go with **3.** then! 

[...]

PR #254, opened 2022-08-20
delete default css

ralfiannor:
Implement #253


matthewmueller:
Thanks for the PR @ralfiannor and for kicking this off! I don't want to remove the default.css without providing an alternative solution. I've wrote my thoughts up in #253. Let me know if you want to drive one of those solutions forward or if you have any other ideas!

framework/view/ssr/svelte.ts
```diff
@@ -58,7 +58,7 @@ const defaultLayout = {
         <html>
           <head>
             <meta charset="utf-8"/>
-            <style>${defaultCSS}</style>
+			<link rel="stylesheet" href="default.css">
```
matthewmueller, line 61: Formatting

framework/view/ssr/svelte.js
```diff
@@ -327,7 +327,7 @@ var defaultLayout = {
         <html>
           <head>
             <meta charset="utf-8"/>
-            <style>${defaultCSS}</style>
+      			<link rel="stylesheet" href="default.css">
```
matthewmueller, line 330: Spacing looks off

internal/cli/create/create_test.go
```diff
@@ -38,6 +38,7 @@ func TestCreateOutsideGoPath(t *testing.T) {
 	is.NoErr(td.Exists("go.sum"))
 	is.NoErr(td.Exists("package.json"))
 	is.NoErr(td.Exists("package-lock.json"))
+	is.NoErr(td.Exists("public/default.css"))
```
matthewmueller, line 41: I think we're missing some test cases still:
1. The scaffolded `public/default.css` gets deleted but views and public files still work as expected.
2. Someone brings their own `public/default.css` and that gets picked up and used.

internal/embedded/embedded.go
```diff
@@ -7,7 +7,19 @@ import (
...
+//go:embed normalize.css
+var normalize string
+
 // Favicon returns the favicon data
 func Favicon() []byte {
 	return favicon
 }
+
+// DefaultCss returns the default css data
+func DefaultCss(css string) string {
```
matthewmueller, line 19: I think we can just use function names like `func Normalize() []byte` or `func Reset() []byte` since it's pretty clear where the context comes from.

matthewmueller (commented):
Thanks for working on this! I'd just like us to clean a few things up and add a few more tests (maybe in framework/public/public_test.go) and then I think we're good to go!

internal/cli/create/create_test.go
ralfiannor: Added the test case in `public_test.go`

framework/controller/controller_test.go
```diff
@@ -1285,7 +1285,6 @@ func TestViewUnnamed(t *testing.T) {
 	is.NoErr(err)
 	is.NoErr(res.DiffHeaders(`
 		HTTP/1.1 200 OK
-		Transfer-Encoding: chunked
```
matthewmueller, line 1288: I'm quite surprised the changes affected these tests. Do you know why you needed to remove this?

internal/embedded/embedded.go
```diff
@@ -7,7 +7,20 @@ import (
...
+
+// NormalizeCss returns the normalize css data
+func NormalizeCss() []byte {
+	return normalize
+}
+
+// EmptyCss reset the default css data
+func EmptyCss() []byte {
+	return []byte("/* No Default CSS Loaded */")
+}
```
matthewmueller, line 26: It'd be better if the view didn't include the `<link />` if a `public/default.css` isn't present since that would avoid a roundtrip for this CSS sheet. 

Then we can also remove this function.

internal/cli/cli.go
```diff
@@ -49,6 +49,7 @@ func (c *CLI) Run(ctx context.Context, args ...string) error {
 		cmd := create.New(cmd, c.in)
 		cli := cli.Command("create", "create a new app")
 		cli.Arg("dir").String(&cmd.Dir)
+		cli.Flag("css", "add a css").String(&cmd.Css).Default("normalize")
```
matthewmueller, line 52: Let's remove this flag for now, I'm not ready to expose it just yet, especially since it's not really possible to select an alternative.

framework/public/public_test.go
```diff
@@ -144,8 +149,20 @@ func TestGetChangeGet(t *testing.T) {
...
+	is.Equal(res.Body().Bytes(), defaultCss)
+
+	// default.css changed
+	defaultCss2 := []byte("/* changed default.css */")
+	td.BFiles["public/default.css"] = defaultCss2
+	is.NoErr(td.Write(ctx))
+	res, err = app.Get("/default.css")
+	is.NoErr(err)
+	is.Equal(200, res.Status())
+	is.Equal(res.Body().Bytes(), defaultCss2)
```
matthewmueller, line 165: It's a bit hard to follow these changes because you've added new conditions to existing tests. 

Let's try to only touch existing tests that need to be updated for the new behavior, then add new tests for the new behavior.

framework/view/ssr/svelte.js
```diff
@@ -327,7 +327,7 @@ var defaultLayout = {
         <html>
           <head>
             <meta charset="utf-8"/>
-            <style>${defaultCSS}</style>
+            <link rel="stylesheet" href="default.css" />
```
matthewmueller, line 330: Do you mind trying to feed the configuration through so you can remove this if `public/default.css` doesn't exist? Let me know if it's difficult and I'll give it a go in this PR.

matthewmueller (changes requested):
Just reviewed. This is getting closer @ralfiannor!

matthewmueller:
I just realized we should have the same treatment for the public/favicon.ico: scaffold a default one, but then give users the option to override it or delete it.

framework/view/ssr/svelte.js
ralfiannor: Will try. I will add the checking file existing in `ssr.go` but will call our logic each page loaded. Is it okay ?

matthewmueller: I think it should be done at compile time actually

framework/controller/controller_test.go
ralfiannor: Reverted. But I still get an error on it while running the unit test. I use `go 1.7` and `node 18.9`

ralfiannor:
> I just realized we should also do the same thing for `public/favicon.ico`: scaffold a default one, but then give users the option to override it or delete it.

Can we do it on another PR?

framework/public/public_test.go
ralfiannor: Reverted. No need to update on `public.go` just update on `ssr.go`

internal/embedded/embedded.go
ralfiannor: Done

internal/cli/cli.go
ralfiannor: Done

matthewmueller (approved):

matthewmueller:
Thanks for continuing to push this forward @ralfiannor! 

Do you know why Go is adding/removing `Transfer-Encoding: chunked`? 

Let's just remove the `Transfer-Encoding: chunked` in the view tests and I think this is good to go!

ralfiannor:
> Do you happen to know the condition in which Go is adding/removing `Transfer-Encoding: chunked`?

Yes, it happen when we added <style></style> the chunked also added

matthewmueller:
Thanks @ralfiannor! I just pulled it down and did some manual testing. Two issues popped up:

1. Welcome page no longer shows up when creating a new project
2. Renaming or removing the `default.css` requires you to restart the `bud` process

Adjusting the global `svelteRuntime` is a bit hacky. If you're up for cleaning this up, a better solution would be to pass this information through the `svelte.gotext` template and then reference that information in the `svelte.ts` file. I'm okay with cleaning this up later if you're not sure how to do this.

Here's a diff with the fixes:

```diff
diff --git a/framework/app/app_test.go b/framework/app/app_test.go
index b5ba644..500a117 100644
--- a/framework/app/app_test.go
+++ b/framework/app/app_test.go
@@ -30,3 +30,26 @@ func TestWelcome(t *testing.T) {
 	is.NoErr(td.Exists("bud/app"))
 	is.NoErr(app.Close())
 }
+
+func TestWelcomeWithPublic(t *testing.T) {
+	is := is.New(t)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	dir := t.TempDir()
+	td := testdir.New(dir)
+	td.Files["public/default.css"] = `* { box-sizing: border-box; }`
+	is.NoErr(td.Write(ctx))
+	cli := testcli.New(dir)
+	is.NoErr(td.NotExists("bud/app"))
+	app, err := cli.Start(ctx, "run")
+	is.NoErr(err)
+	res, err := app.Get("/")
+	is.NoErr(err)
+	is.Equal(res.Status(), 200)
+	is.In(res.Body().String(), "Hey Bud")
+	is.In(res.Body().String(), "Hey Bud") // should work multiple times
+	is.Equal(app.Stdout(), "")
+	is.Equal(app.Stderr(), "")
+	is.NoErr(td.Exists("bud/app"))
+	is.NoErr(app.Close())
+}
diff --git a/framework/view/ssr/ssr.go b/framework/view/ssr/ssr.go
index 7116bcd..8bab21c 100644
--- a/framework/view/ssr/ssr.go
+++ b/framework/view/ssr/ssr.go
@@ -50,9 +50,10 @@ type Compiler struct {
 
 func (c *Compiler) Compile(ctx context.Context, fsys budfs.FS) ([]byte, error) {
 	dir := c.module.Directory()
-
 	if existCss := vfs.Exist(fsys, "public/default.css"); nil == existCss {
 		svelteRuntime = strings.Replace(svelteRuntime, `<!-- default css -->`, `<link rel="stylesheet" href="default.css">`, 1)
+	} else {
+		svelteRuntime = strings.Replace(svelteRuntime, `<link rel="stylesheet" href="default.css">`, `<!-- default css -->`, 1)
 	}
 
 	result := esbuild.Build(esbuild.BuildOptions{
diff --git a/framework/web/loader.go b/framework/web/loader.go
index 5f0d149..e4695b8 100644
--- a/framework/web/loader.go
+++ b/framework/web/loader.go
@@ -52,7 +52,10 @@ func (l *loader) Load() (state *State, err error) {
 	l.imports.AddNamed("webrt", "github.com/livebud/bud/framework/web/webrt")
 	l.imports.AddNamed("router", "github.com/livebud/bud/package/router")
 	// Show the welcome page if we don't have controllers, views or public files
-	if len(exist) == 0 {
+	//
+	// TODO: welcome should be part of the default controller, then we can support
+	// a simple static file server without generating the welcome page.
+	if len(exist) == 0 || (len(exist) == 1 && exist["bud/internal/app/public/public.go"]) {
 		l.imports.AddNamed("welcome", "github.com/livebud/bud/framework/web/welcome")
 		state.ShowWelcome = true
 		state.Imports = l.imports.List()

```

matthewmueller:
Any chance of wrapping this up @ralfiannor ?

PR #259, opened 2022-08-26
Issue/171

jfmario:
This implements validation by calling `Validate` when appropriate. There are still several other things to do, so this should be  a draft. I am curious what you think of this approach so far.

### DIFF
--- a/framework/controller/controller.gotext
+++ b/framework/controller/controller.gotext
@@ -69,6 +69,19 @@ func ({{$action.Short}} *{{ $.Pascal }}{{$action.Pascal}}Action) handler(httpRes
 			JSON: response.Status(400).Set("Content-Type", "application/json").JSON(map[string]string{"error": err.Error()}),
 		}
 	}
+	// validate input
+	{{- if (index $action.Params 0).HasValidate }}
+	// single input struct has validate method
+	if err := in.Validate(); err != nil {
+		return &response.Format{
+			{{- if ne $action.Method "GET" }}
+			HTML: response.Status(http.StatusSeeOther).RedirectBack(httpRequest.URL.Path),
+			{{- end }}
+			JSON: response.Status(400).Set("Content-Type", "application/json").JSON(map[string]string{"error": err.Error()}),
+		}
+	}
+	{{- end }}
+	
 	{{- end }}
 	{{- with $provider := $action.Provider }}
 	controller, err := {{ $provider.Name }}(
--- a/framework/controller/controller_test.go
+++ b/framework/controller/controller_test.go
@@ -2792,3 +2792,38 @@ func TestIndexControllerWithRootIndexAction(t *testing.T) {
 	`))
 	is.NoErr(app.Close())
 }
+
+// Asserts that a root index action is compatible with an index
+// controller. Also asserts that a posts index action is compatible
+// with a posts/index controller.
+func TestInputValidateMethod(t *testing.T) {
+	is := is.New(t)
+	ctx := context.Background()
+	dir := t.TempDir()
+	td := testdir.New(dir)
+	// root controller with input struct to validate
+	td.Files["controller/controller.go"] = `
+		package controller
+		import "errors"
+		type Controller struct{}
+		type Input struct{}
+		func (in *Input) Validate() error {
+			return errors.New("")
+		}
+		func (c *Controller) Index(in *Input) []string {
+			return []string{}
+		}
+	`
+	is.NoErr(td.Write(ctx))
+	cli := testcli.New(dir)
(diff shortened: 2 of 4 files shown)


framework/controller/controller.gotext
```diff
@@ -69,6 +69,19 @@ func ({{$action.Short}} *{{ $.Pascal }}{{$action.Pascal}}Action) handler(httpRes
 			JSON: response.Status(400).Set("Content-Type", "application/json").JSON(map[string]string{"error": err.Error()}),
 		}
 	}
+	// validate input
+	{{- if (index $action.Params 0).HasValidate }}
```
matthewmueller, line 73: I've been trying to avoid adding logic like into the templates because the templates are difficult to understand as-is. 

I'd create a method on the state that could be used, so it's more like `{{- if $action.Params.HasValidate }}`.

I'm not sure if action params has this already, but I'm pretty sure the action results does already, so it would be a similar change here.

framework/controller/controller.gotext
```diff
@@ -69,6 +69,19 @@ func ({{$action.Short}} *{{ $.Pascal }}{{$action.Pascal}}Action) handler(httpRes
...
+	// single input struct has validate method
+	if err := in.Validate(); err != nil {
+		return &response.Format{
+			{{- if ne $action.Method "GET" }}
+			HTML: response.Status(http.StatusSeeOther).RedirectBack(httpRequest.URL.Path),
```
matthewmueller, line 78: The big question is going to be what happens with the error message? Right now if a form fails, there's no feedback sent back to the user. 

[Flash messages](https://api.rubyonrails.org/classes/ActionDispatch/Flash.html) are typically the solution to this where you store a value in a cookie that gets removed after it's pulled out. I think the ultimate solution is to have the ability to add [custom flash messages via the session](https://laravel.com/docs/9.x/session#flash-data), but if there's any temporary solution to get this working where you set the cookie, pull it out and pass to the view, I'd be open to pursuing that. Here's what I've thought about so far with sessions: https://github.com/livebud/bud/discussions/56.

Otherwise maybe we only enable this for JSON? Does your use case fit with that limitation?

matthewmueller (commented):
Thanks for kicking this off! I've just reviewed. I really like where you started with this PR because it would remain generic while still allowing you to stick an existing validator library in there and do something like:

```go
type Input struct {
  name string `validate:len>5`
}

var validate = validator.New()

func(i *Input) Validate() error {
  return validate.Struct(i)
}
```

_(not sure on exact syntax)_

framework/controller/controller.gotext
```diff
@@ -69,6 +69,19 @@ func ({{$action.Short}} *{{ $.Pascal }}{{$action.Pascal}}Action) handler(httpRes
 			JSON: response.Status(400).Set("Content-Type", "application/json").JSON(map[string]string{"error": err.Error()}),
 		}
 	}
+	// validate input
+	{{- if (index $action.Params 0).HasValidate }}
+	// single input struct has validate method
+	if err := in.Validate(); err != nil {
```
matthewmueller, line 75: What are your thoughts on calling it `Validate` vs `Valid`? IMO `Validate` sounds more like an action that would return an error, but `Valid` is shorter.

jfmario:
Ok, I'll clean this up.

Also, I just re-read the discussion here: https://github.com/livebud/bud/discussions/15

`Validate()` (or `Valid()`) also has to work on types besides structs, so I'll re-work this a bit.

framework/controller/controller.gotext
jfmario: Yea, I think flash messages should be the behavior here. I'm thinking to add some other stuff to this PR but then put in on pause until we have sessions and something like flash messages.

framework/controller/controller.gotext
jfmario: I actually like `Valid()` now that you bring it up. I was doing `Validate()` because that was what was in the RFC 15. I don't feel strongly either way.

matthewmueller: Oh right, let's go with `Valid()` then!

framework/controller/controller.gotext
matthewmueller: Cool, sounds good! I'm going to get back to feature development very soon. Just trying to wrap up generator caching to allow for more expensive generators to be able to run (e.g. generators that call `go run`)

jfmario:
I changed the name to `Valid()` but this is still a draft and depends on sessions, I think.

matthewmueller:
Yah, might be best to wait for sessions. In the meantime, you can always add the validate function to the method body. 

I'll try and get sessions added soon!

PR #261, opened 2022-08-26
+1 to port if necessary

jfmario:
This PR will have the web server keep incrementing the port until it finds one that it can bind to without error.


matthewmueller:
Awesome, thanks for this! We'll just need a test or two before we accept this. I typically add these types of tests to https://github.com/livebud/bud/blob/298b70593910ec08fae99fe9a700fc728021e4fa/framework/app/app_test.go.

To test this you can probably just before starting the server listen on `:3000`:

```go
	listener, err := socket.Listen(":3000")
	if err != nil {
		return nil, nil, err
	}
```

Then test that the server was bound to :3001

jfmario:
@matthewmueller Do you have any insight into why sending a GET request to `localhost:3000` and to `localhost:3001` both fail? During testing, does the app not actually run on a socket?

matthewmueller:
Hey @jfmario! I had a look at the code and you're right, the test suite uses `:0` to avoid addresses being bound. You can find that logic here: https://github.com/livebud/bud/blob/5f9309e5db27cf4edd45c3f19a0689c0465813b7/internal/cli/testcli/testcli.go

In attempting to add tests to this PR, I realized it would be easier to add a method and test in the socket package instead. I found this additionally preferable because it reduces the amount of code in an already complex `run` package.

Closing in favor of https://github.com/livebud/bud/pull/271.

Thanks for your effort though! Your work sped up this PR considerably.

PR #264, opened 2022-08-28
don't trigger generators with stat() or filler dir's readdir()

matthewmueller:
This PR reduces the scope of the virtual filesystem in order to reduce the number of calls to the generator functions.

After this PR:

- `stat.Size()` will always return -1 for generators
- `stat.ModTime().IsZero()` will always be true for generators
- `fs.Stat(genfs, filepath)` won't call the underlying generate function
- `fs.ReadDir(genfs, somedir)` won't call the underlying generate function of the dir entries

This is a tradeoff, but I think it's going to be worth it. It will be more clear when generators run at the expense of being a less accurate filesystem. Surprisingly only a few of the tests failed and none of the framework tests failed after this change.

### DIFF
--- a/internal/virtual/dir.go
+++ b/internal/virtual/dir.go
@@ -1,16 +1,13 @@
 package virtual
 
 import (
+	"encoding/json"
 	"io"
 	"io/fs"
-	"time"
 )
 
 type Dir struct {
 	Path    string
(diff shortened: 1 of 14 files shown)


matthewmueller:
Closing in favor of #268.

PR #277, opened 2022-09-05
Filesystem caching

matthewmueller:
One of the great things about building a framework and having control over the build system is that we can implement very good caching. So far the caching has been limited to not re-running generators within a single sync, but when you call `bud run` there's a sync every time you change a Go file. 

Syncing is still quite fast, but I'd like to be able to spawn subcommands during sync to enable features like custom generators. Restarting processes each sync is time-consuming.  Instead we'll only restart processes when dependencies of the custom generators change (almost never).

Caching is quite hard to get right, so we may run into some stale builds along the way, but I'm confident these issues can be ironed out over time. 

Next steps:
- [ ] Each change should be tested against directed acyclic graph (dag) of globs
   This will allow the controller generator to react to a new file in `view/**`
- [ ] Get the controller tests passing
- [ ] Manual QA


matthewmueller:
Closing in favor of: #281

Issue #279, opened 2022-09-08
How do I use UI components?


JanGordon:
Hello, I am very interested in this project but I can't work out where I am going wrong here.
I don't understand how to import ui components from svelte files.
I have a folder in the base of the project named "ui" and a file named "Test.svelte" in it.
If I try to import the component in the main index.svelte of the page like this:
`import Test from "ui/Test.svelte"`
I get this error:
` You can mark the path "ui/Test.svelte" as external to exclude it from the
  bundle, which will remove this error.`

I also get the same error when I try and import an asset from the public directory like so:
`import image from "public/image.png"`

matthewmueller:
Hey @timmyjan46, thanks for your interest in Bud!

I'm surprised that `import Test from "ui/Test.svelte"` doesn't work. If you have a spare moment, creating a failing test isn't too difficult and would speed up the fix: https://github.com/livebud/bud/blob/2b3cf758554bdb6fa385129955b4ed50637f50bd/framework/view/view_test.go#L361. No pressure, either way I'll give this a shot soon.

Importing non-JS/TS/Svelte assets doesn't work yet and I'm not sure if that's something worth adding. What do you expect `image` to be in `import image from "public/image.png"`?

timmyjan46:
Thanks for answering so quickly!

I was expecting `import image from "public/image.png"` to give me a URL as it does when using vite to build. You can use the image variable for and image source.

I can try and work out how to add a test but I'm not sure if I'll succeed but I'll give it a go.

matthewmueller:
> I was expecting import image from "public/image.png" to give me a URL as it does when using vite to build. You can use the image variable for and image source.

Oh got it, I'll have a look at how vite does it. I'm on the fence about importing assets via JS because sometimes I could see people expecting a data-uri. I do see the convenience in this and maybe I can make it pluggable.

> I can try and work out how to add a test but I'm not sure if I'll succeed but I'll give it a go.

Saw your PR, thanks!

JanGordon:
[...]

PR #280, opened 2022-09-09
Add svelte import test

JanGordon:
A test that checks if you can import a svelte component from another svelte file.
 
I hope I did the right thing that you asked for [here](https://github.com/livebud/bud/issues/279)

### DIFF
--- a/framework/view/view_test.go
+++ b/framework/view/view_test.go
@@ -414,3 +414,43 @@ func TestSvelteImportFromNodeModule(t *testing.T) {
 	is.In(res.Body().String(), "<time datetime=\"2022-07-19 10:19:00\">Jul 19, 2022</time>")
 	is.NoErr(app.Close())
 }
+
+func TestSvelteImportFromSvelte(t *testing.T) {
+	is := is.New(t)
+	ctx := context.Background()
+	dir := t.TempDir()
+	td := testdir.New(dir)
+	td.NodeModules["svelte"] = versions.Svelte
+	td.NodeModules["svelte-time"] = "*"
+	td.NodeModules["livebud"] = "*"
+	td.Files["controller/controller.go"] = `
+		package controller
+		type Controller struct {}
+		func (c *Controller) Index() string { return "" }
+	`
+	td.Files["ui/Time.svelte"] = `
+		<h1>The Time</h1>
+	`
+	td.Files["view/index.svelte"] = `
+	 	<script>
+			import Time from "../../ui/Time.svelte"
+		</script>
+		<Time/>
+	`
+	// only works with "import Time from "../ui/Time.svelte" however this is doesnt work with views in nested directories
+
+	is.NoErr(td.Write(ctx))
+	cli := testcli.New(dir)
+	app, err := cli.Start(ctx, "run")
+	is.NoErr(err)
+	defer app.Close()
+	res, err := app.Get("/")
+	is.NoErr(err)
+	is.NoErr(res.DiffHeaders(`
+		HTTP/1.1 200 OK
+		Transfer-Encoding: chunked
+		Content-Type: text/html
+	`))
+	is.In(res.Body().String(), "<h1>The Time</h1>")
+	is.NoErr(app.Close())
+}


framework/view/view_test.go
```diff
@@ -391,3 +391,43 @@ func TestSvelteImportFromNodeModule(t *testing.T) {
...
+		package controller
+		type Controller struct {}
+		func (c *Controller) Index() string { return "" }
+	`
+	td.Files["ui/Time.svelte"] = `
+		<h1>The Time</h1>
+	`
+	td.Files["view/index.svelte"] = `
+	 	<script>
+			import Time from "ui/Time.svelte"
```
matthewmueller, line 413: Oh, this is currently intentional. Paths that don't have `./` or `../` are considered node_modules. I'd definitely accept a change where we add support for `~/` to reference the project root, but for now, this is working as intended.

framework/view/view_test.go
```diff
@@ -391,3 +391,43 @@ func TestSvelteImportFromNodeModule(t *testing.T) {
...
+		</script>
+		<Time/>
+	`
+	// only works with "import Time from "../ui/Time.svelte" however this is doesnt work with views in nested directories
```
matthewmueller, line 417: Do you mean that `../ui/time/Time.svelte` wouldn't work?

matthewmueller:
Awesome, this is very helpful @timmyjan46! I just commented inline.

framework/view/view_test.go
timmyjan46: Oh, ok, sorry for my misunderstanding of your last comment. However, I don't understand how you are meant to access files in the ui folder from a view. In the [documentation](https://denim-cub-301.notion.site/Hey-Bud-4d81622cc49942f9917c5033e5205c69#326e9e129af44fffaad0156f225ce3f2) it has the ui folder but doesn't show how to import from it:

$YOUR_APP
└─ ui
│  ├─ Calendar.svelte
│  └─ Popover.svelte
└─ view
   └─ users
      ├─ Avatar.svelte
      ├─ Card.svelte
      ├─ index.svelte
      └─ show.svelte

It then has a code snippet showing you how to import the Card.svelte but not the Calendar or Popover:

`<script>
  import Card from './Card.svelte'
  export let users = []
</script>

{#each users as user}
<Card {user} />
{/each}`

I'm sorry for the trouble and I hope I am not using bud in the wrong way.

framework/view/view_test.go
timmyjan46: No, I mean that does work, but only in the root of the view folder and not in any nested folders within it

framework/view/view_test.go
matthewmueller: Hey @JanGordon , you'd be able to import it with the following:

```svelte
import Calendar from "../ui/Time.svelte"
```

framework/view/view_test.go
matthewmueller: In a nested directory, does the following not work `../../ui/time/Time.svelte`? It should work with any relative path.

framework/view/view_test.go
JanGordon: But I don't think that would work for any route. It would only work if the file is directly in the views folder and not nested in another folder, right? I'm a bit confused because I would have thought you would have to import components a lot. Am I trying to do things wrong? Or is there a way to reference the root of the project from anywhere?

framework/view/view_test.go
JanGordon: No, doesn't work. It just returns an error: could not resolve ../../ui/Time.svelte. I definitely have the Time.svelte file in the UI folder.

framework/view/view_test.go
JanGordon: But I don't think that would work for any route. It would only work if the file is directly in the views folder and not nested in another folder, right? I'm a bit confused because I would have thought you would have to import components a lot. Am I trying to do things wrong? Or is there a way to reference the root of the project from anywhere?

framework/view/view_test.go
matthewmueller: Ah cool, can you update the PR to demonstrate that? I'll fix this ASAP :)

JanGordon: I've just done that

matthewmueller:
Hey @JanGordon, just updated your PR in another PR: https://github.com/livebud/bud/pull/283. 

The test contains the expected relative path behavior, so you'll find a working example in that PR. Thanks for kicking off this PR!

PR #288, opened 2022-09-18
framework/transform: initial transform generator

matthewmueller:
This PR adds a transform generator for transforming one type to another type (e.g. `index.svelte` => `index.ssr.js`).

You can create your own transforms by adding packages in the `transform/` directory:

```
transform/
  tailwind/
    tailwind.go
  markdoc/
    markdoc.go
```

And inside these packages, you'd have the following:

```go
package tailwind

import (
  "github.com/livebud/bud/framework/transform/transformrt" // import will likely change
  "bytes"
)

type Transform struct {
  Tailwind *tailwind.Compiler
}

func (t *Transform) SvelteToSvelte(file *transformrt.File) error {
  css, err := t.Tailwind.Compile(file.Data)
  if err != nil {
    return err
  }
  file.Data = []byte("<style>" + css + "</style>" + string(code))
  return nil
}
```

This would translate a `.svelte` file back into a `.svelte` file, but perhaps turn the tailwind classes into a `<style/>` tag.

One of the neat things about this implementation is that the transforms can depend on each other. For example, if you defined:

- MdToSvelte (e.g. `.md` to `.svelte`)
- SvelteToSsrJS (e.g. `.svelte` to `.ssr.js`)
- SvelteToDomJS (e.g. `.svelte` to `.dom.js`)
- SvelteToSvelte (e.g. `.svelte` to `.svelte`)

And you made the following request to the virtual filesystem: `bud/transform/view/index.svelte..ssr.js`, it would run the following transformation pipeline:

1. MdToSvelte
2. SvelteToSvelte
3. SvelteToSsrJS

Currently I'm planning on exposing the transform generator in the `view` and `public` packages. You can think of this like an internal service that can be added to other generators that need to transform files.

**Todos**
- [x] Refactor `generator` and `transform` generators to spawn a single executable for executing custom application code.
- [ ] Replace current internal transform implementation with this generator (transform2)


matthewmueller:
Replaced by https://github.com/livebud/bud/pull/320

PR #334, opened 2022-11-28
feature: automatically launch browser on `bud run`

### LINKED ISSUE #333: feature suggestion: `bud run` should launch browser
When launching the dev server, I have to manually open a browser tab and load the URL.  It would be cool if it called out to the OS and simply told it to open the dev server URL.  The quick and dirty (on macOS) is to shell out to `open $URL` but I imagine there's a go module somewhere that does this.

sneak:
fixes #333

### DIFF
--- a/framework/framework.go
+++ b/framework/framework.go
@@ -4,9 +4,10 @@ import "io"
 
 // Flag is used by many of the framework generators
 type Flag struct {
-	Embed  bool
-	Minify bool
-	Hot    bool
+	Embed        bool
+	Minify       bool
+	Hot          bool
+	Noautolaunch bool
 
 	// Comes from *bud.Input
 	Stdin  io.Reader
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -63,6 +63,7 @@ func (c *CLI) Run(ctx context.Context, args ...string) error {
 		cli.Flag("embed", "embed assets").Bool(&cmd.Flag.Embed).Default(false)
 		cli.Flag("hot", "hot reloading").Bool(&cmd.Flag.Hot).Default(true)
 		cli.Flag("minify", "minify assets").Bool(&cmd.Flag.Minify).Default(false)
+		cli.Flag("noautolaunch", "don't autolaunch browser").Bool(&cmd.Flag.Noautolaunch).Default(false)
 		cli.Flag("listen", "address to listen to").String(&cmd.Listen).Default(":3000")
 		cli.Run(cmd.Run)
 	}
--- a/internal/cli/run/run.go
+++ b/internal/cli/run/run.go
@@ -8,8 +8,6 @@ import (
 	"path/filepath"
 	"time"
 
-	"golang.org/x/sync/errgroup"
-
 	"github.com/livebud/bud/framework"
 	"github.com/livebud/bud/framework/web/webrt"
 	"github.com/livebud/bud/internal/bfs"
@@ -26,6 +24,8 @@ import (
 	"github.com/livebud/bud/package/log"
 	"github.com/livebud/bud/package/socket"
 	"github.com/livebud/bud/package/watcher"
+	"github.com/pkg/browser"
+	"golang.org/x/sync/errgroup"
 )
 
 // New command for bud run.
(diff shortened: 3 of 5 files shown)


internal/cli/run/run.go
```diff
@@ -150,6 +150,15 @@ func (c *Command) Run(ctx context.Context) (err error) {
 	// Start the internal app server
 	eg.Go(func() error { return appServer.Run(ctx) })
 	// Wait until either the hot or web server exits
+
+	// Launch browser unless we're told not to
+	if !c.Flag.Noautolaunch { // if not not autolaunch, i.e. if autolaunch
+		go func() {
+			time.Sleep(1 * time.Second)
```
matthewmueller, line 157: Any reason for the sleep? At this point, the socket is open so it will hang.

internal/cli/run/run.go
```diff
@@ -150,6 +150,15 @@ func (c *Command) Run(ctx context.Context) (err error) {
 	// Start the internal app server
 	eg.Go(func() error { return appServer.Run(ctx) })
 	// Wait until either the hot or web server exits
+
+	// Launch browser unless we're told not to
+	if !c.Flag.Noautolaunch { // if not not autolaunch, i.e. if autolaunch
+		go func() {
```
matthewmueller, line 156: I don't think the goroutine is necessary.

internal/cli/run/run.go
```diff
@@ -150,6 +150,15 @@ func (c *Command) Run(ctx context.Context) (err error) {
...
+			time.Sleep(1 * time.Second)
+			_ = browser.OpenURL("http://" + webln.Addr().String())
```
matthewmueller, line 158: Should check the error

internal/cli/cli.go
```diff
@@ -63,6 +63,7 @@ func (c *CLI) Run(ctx context.Context, args ...string) error {
 		cli.Flag("embed", "embed assets").Bool(&cmd.Flag.Embed).Default(false)
 		cli.Flag("hot", "hot reloading").Bool(&cmd.Flag.Hot).Default(true)
 		cli.Flag("minify", "minify assets").Bool(&cmd.Flag.Minify).Default(false)
+		cli.Flag("noautolaunch", "don't autolaunch browser").Bool(&cmd.Flag.Noautolaunch).Default(false)
```
matthewmueller, line 66: Let's go with `open` defaulting to false

framework/framework.go
```diff
@@ -4,9 +4,10 @@ import "io"
 
 // Flag is used by many of the framework generators
 type Flag struct {
-	Embed  bool
-	Minify bool
-	Hot    bool
+	Embed        bool
+	Minify       bool
+	Hot          bool
+	Noautolaunch bool
```
matthewmueller, line 10: Open

matthewmueller (changes requested):
Thanks for opening this PR!

sneak:
Thanks for the review and feedback- I will update the PR soon.

internal/cli/run/run.go
sneak: this is the only guidance here i don't understand. why? i literally can't think of anything that we'd do differently if it fails or not.  it's an optional convenience feature...

matthewmueller: My take is that if a user explicitly requested `--open` we should open or fail trying.

internal/cli/cli.go
```diff
@@ -63,6 +63,7 @@ func (c *CLI) Run(ctx context.Context, args ...string) error {
 		cli.Flag("embed", "embed assets").Bool(&cmd.Flag.Embed).Default(false)
 		cli.Flag("hot", "hot reloading").Bool(&cmd.Flag.Hot).Default(true)
 		cli.Flag("minify", "minify assets").Bool(&cmd.Flag.Minify).Default(false)
+		cli.Flag("open", "open browser on dev server startup").Bool(&cmd.Flag.OpenBrowser).Default(false)
```
matthewmueller, line 66: Ah sorry, I might have missed this in the previous review but Flag is meant for global options that need to get passed into the framework. We should treat `OpenBrowser` like Listen below and stick it on the run command.

matthewmueller (commented):
Looks a lot better, thanks the update!

Issue #332, opened 2022-11-27
feature idea: detect git short version hash during build and embed into binary


sneak:
In my apps I usually do something like:

```
var Version   string

func main() {
	os.Exit(httpserver.Run(Version))
}
```

Then at build time (in my Makefile) I do this:

```
VERSION := $(shell git describe --always --dirty=-dirty)
GOLDFLAGS += -X main.Version=$(VERSION)
GOFLAGS := -ldflags "$(GOLDFLAGS)"
```

This embeds the version string into the app so that I can access it, which is useful for me because then in production I can access this value and put it in my footers and error messages so that I know precisely which commit is running.

It would be cool if `bud build` could detect if it's building out of a git repo, and embed the git shorthash into the built binary in this way.  It seems that you could skip the `ldflags` part and simply "hardcode" the git shorthash into the main.go via the main.gotext template at build filesystem generation time.

Anyway, just an idea, and I'm happy to take a crack at making a patch for this when I have some time, I just wanted to write it down while it's in my head.

syke99:
Hey @matthewmueller I wouldn't mind knocking this out. 1 should be more than simple enough. However, do you have any input for tacking 2? It's been a minute since I've looked at the bud codebase, so I gave it a cursory glance and noticed `internal/cli/build/build.go` directory. Gave that a quick look and noticed `Command` has the field `Flag` of `*framework.Flag` type. Would this be the appropriate jumping off point or do you have any pointers of where I should look if this isn't the right spot?

PR #340, opened 2022-12-14
feature idea: detect git short version hash during build and embed into binary

syke99:
This PR addresses the first proposed solution to [this](https://github.com/livebud/bud/issues/332) issue. The second proposed solution is not implemented in this PR, but can be added with further information provided.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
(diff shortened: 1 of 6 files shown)


matthewmueller:
Hey @syke99 thanks for working on this! 

I'm a bit confused by the PR though. It's not really connected to Bud in anyway. You'd need to add the `-ldflags` to the app command. Is this still a draft PR?

I'm also thinking we can simplify this buildinfo package quite a bit, pretty sure something like: https://github.com/livebud/bud/blob/d243eedcb7e71b0c323c4771397c051a7872bf6c/internal/versions/versions.go#L4 would suffice.

syke99:
Hey @matthewmueller, I'm so sorry. Yes, this was supposed to just be a draft PR for now. I was expecting to continue being able to work on it, but I've had some personal stuff come up that has been taking precedence, so I've had to step away. And you're right, that does seem like it would suffice for what I had added. Again, my apologies!!

PR #344, opened 2022-12-22
Prep for new budfs

matthewmueller:
Little PR to shuffle packages around.


almas1992:
Suggest changing `github.com/livebud/bud/internal/dag2` to `github.com/livebud/bud/internal/dag/v2`, and file path like this 
```bash
internal
└── dag
    ├── dag.go
    ├── dag_test.go
    └── v2
        ├── dag.go
        └── dag_test.go
```

matthewmueller:
@almas1992 thanks for the suggestion! In this case, it's a temporary internal package that's going to be removed once the refactor is complete, so it's better to just rename it for now.

PR #364, opened 2023-01-23
package/di: improve dependency injection with type aliases

matthewmueller:
In preparation for adding the transpiler generator, I stumbled across a bug with dependency injection with type aliases. This PR fixes that bug.


package/di/di_test.go
```diff
@@ -2466,6 +2466,7 @@ func TestV8Alias(t *testing.T) {
 }
 
 func TestAliasBack(t *testing.T) {
+	t.Skip("TODO: finish up after")
```
matthewmueller, line 2469: Fix this before merging

PR #378, opened 2023-02-09
build: support Go v1.20 and add to CI checks

wheinze:
Add support for Go 1.20 to GitHub action.


matthewmueller:
Thanks for helping with this @wheinze! Do you have any idea what's going on with the failures?

matthewmueller:
I haven't yet upgraded. A bit of a slow adopter 😅 I'll give upgrading a
try this coming weekend.

Unfortunately not sure what the error is and was hoping you knew 😄

wheinze:
Nope, :laughing: , but we'll make it ;)

staticcheck issues seem to be resolved by using the proper released version of it, as stated per [docs](https://github.com/livebud/bud/pull/378). I [changed](https://github.com/livebud/bud/pull/378/commits/10493a96578646e1778749576c604ddd58bbbecb) it and those issues related to staticcheck combined wiith Go 1.20 do not appear anymore, resulting in the same issues with the failing tests now: https://github.com/livebud/bud/actions/runs/4171476823/jobs/7221455414

The drawback here seems to be, that older Go versions are not supported anymore:
- Go 1.17: https://github.com/livebud/bud/actions/runs/4171476823/jobs/7221455061
- Go 1.18: https://github.com/livebud/bud/actions/runs/4171476823/jobs/7221455172

staticcheck requires at least Go 1.19. Which brings me to the question if we should also lift requirements for bud, or differentiate on this in CI and use older versions of staticcheck for Go <1.19. @matthewmueller What do you think?

As updating your global Go version, most likely, will break your local env as well, you might consider giving [asdf-vm](https://asdf-vm.com/) a shot. I provided a [PR](https://github.com/livebud/bud/pull/381) for main with Go 1.19 and would add a proper .tool-versions file here.

matthewmueller:
Progress! I'd personally like to keep older Go versions working over time. I have been considering using Generics which would bump it up to 1.18 I think.

It looks like we can target specific versions of Go with: https://staticcheck.io/docs/running-staticcheck/cli/#go. Maybe we can bump staticcheck, but then use the `--go` flag to swap the checks depending on which Go version is running in CI?

.tool-versions
```diff
@@ -9,5 +9,5 @@
 # a more complete list, review the GitHub workflow to see the versions tested
 # in CI.
 
-golang 1.19.3
+golang 1.20.2
```
wheinze, line 12: @matthewmueller Should we already increase the default Go version with this PR as its working right now? What do you think?

wheinze:
> Progress! I'd personally like to keep older Go versions working over time. I have been considering adopting generics in Bud which would bump it up to 1.18, but that seems like a worthwhile reason when the time comes.

Fair point! Yep, we should not bump Bud's requirements because of the used tooling that demands it , but rather only do this when needed for Bud itself :+1: 

 
> It looks like we can target specific versions of Go with: https://staticcheck.io/docs/running-staticcheck/cli/#go. Maybe we can bump staticcheck, but then use the `--go` flag to swap the checks depending on which Go version is running in CI?


Nice catch! As staticcheck provides a GitHub action, I used that one to integrate more native into GtHub's CI. The action also differentiates on the version of staticcheck that is used as they only provide the last two Go versions with a release. The staticcheck job runs fine with this approach ;)

framework/generator/generator_test.go
```diff
@@ -180,7 +180,7 @@ func TestSyntaxError(t *testing.T) {
 	cli := testcli.New(td.Directory())
 	res, err := cli.Run(ctx, "build", "--embed=false")
 	is.True(err != nil)
-	is.In(err.Error(), `exit status 2`)
+	is.In([]string{`exit status 1`, `exit status 2`}, err.Error()) // The exit status code changed in Go v1.20 from "2" to "1".
```
matthewmueller, line 183: Did you see this documented somewhere online? 

I had a look at the release notes, but didn't spot anything. Also grepped the code, but no `os.Exit(2)`

.tool-versions
matthewmueller: Yep sounds good!

.github/workflows/ci.yml
```diff
@@ -75,3 +78,35 @@ jobs:
...
+          cache-key: ${{ matrix.go }}
+          working-directory: ${{ matrix.dir }}
+      - uses: dominikh/staticcheck-action@v1.3.0
+        if: contains(fromJSON('["1.17", "1.18"]'), matrix.go)
+        with:
+          version: '2022.1'
+          install-go: false
+          min-go-version: ${{ matrix.go }}
+          cache-key: ${{ matrix.go }}
+          working-directory: ${{ matrix.dir }}
```
matthewmueller, line 112: Just for my understanding, this was pulled out of the Makefile so we could run different staticcheck binaries (e.g. `2022.1` and `2023.1`) for different versions of Go?

matthewmueller (commented):
Sorry for the delay on this and thank you so much for working on this Waldemar!

framework/generator/generator_test.go
wheinze: Sadly no, this "fix" is only based on empirical analysis of Go-version-based behavior :see_no_evil: 

Would like to have this backed by objective facts as well, but could not yet find any. Did also seek through the release notes as well, but could not find any changes regarding this.

Where exactly did you grep for `os.Exit(2)`? 
Quickly grepped through go1.19.7 and go1.20.2 sources which at least gave me a different count sfor it ;) (I would assume  Go <= 1.19 to return with `2` and Go >= 1.20 to return with `1`)

.github/workflows/ci.yml
wheinze: Yeah, it's done to integrate CI more GitHub-agnostic so we could leverage more of its feature like parallelization and visualization of jobs, and having a well defined set of supported Go-versions in that corresponding matrix for maintenance.

I also kept the target `go.staticcheck` in the Makefile for non-CI usage.

Technically, this would also be possible to achieve via changes in `Makefile` of course, but it still is more of an individual solution while GitHub-actions might be what other contributors are already used to work with so it lowers the initial barrier to get involved into contributing to `Bud`.

matthewmueller:
Merged, thanks so much for your continued help @wheinze!

PR #381, opened 2023-02-14
build: add support for asdf-vm

wheinze:
Add support for asdf-vm which also makes switching Go versions in local development setup easy.

Especially useful when debugging issues related to Go version updates (see also #378)


wheinze:
I successfully ran `make ci.ubuntu` in this context.

matthewmueller:
Hey @wheinze, I hadn't heard of `asdf` before. I just tested it out locally. 

My only gripe is that I wish it behaved more like `direnv` where when you left the directory it unloaded the binaries. Otherwise it's fantastic, thanks for bringing this tool to my attention!

wheinze:
Actually there's a pretty neat integration with `direnv` available: https://github.com/asdf-community/asdf-direnv ;)

Will provide a proper PR to make use of it in Bud.

PR #384, opened 2023-02-22
package/esb: initial esbuild package rewrite

matthewmueller:
This PR adds the `esb` (short for esbuild) package to Bud. The purpose of this package is to provide common presets that viewers can use and extend in #383. 

I'm also taking this opportunity to try and move esbuild fully to a virtual filesystem, so it's much easier to test different combos.


matthewmueller:
Note: Going to merge now and fix issues as they come up while working on the viewer.

PR #393, opened 2023-03-12
build: add `direnv` support

wheinze:
As [promised](https://github.com/livebud/bud/pull/381#issuecomment-1437730820), this PR adds support for `direnv` to work together with `asdf` seamlessly.

To make use of it, [direnv](https://direnv.net/) and [asdf-direnv](https://github.com/asdf-community/asdf-direnv) should be set up.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use asdf
--- a/.envrc
+++ b/.envrc
@@ -1,1 +1,3 @@
-use asdf
+if has asdf; then
+    use asdf
+fi
--- a/.tool-versions
+++ b/.tool-versions
@@ -5,6 +5,11 @@
 # We'll use `asdf` during development to easily test between
 # different versions.
 #
+# We recommend to use `direnv` (https://direnv.net/) alongside with `asdf`
+# to load/unload project-specific environments when changing directories.
+# `asdf-direnv`(https://github.com/asdf-community/asdf-direnv) lets play
+# them together nicely.
+#
 # The versions below are not the only valid versions that Bud supports. To see
 # a more complete list, review the GitHub workflow to see the versions tested
 # in CI.


.envrc
```diff
@@ -0,0 +1 @@
+use asdf
```
matthewmueller, line 1: Ahhh, I get it now. Interesting! 

Does this work if asdf isn't installed but direnv is?

.tool-versions
```diff
@@ -5,6 +5,11 @@
 # We'll use `asdf` during development to easily test between
 # different versions.
 #
+# We recommend to use `direnv` (https://direnv.net/) alongside with `asdf`
+# to load/unload project-specific environments when changing directories.
+# `asdf-direnv`(https://github.com/asdf-community/asdf-direnv) lets play
+# them together nicely.
```
matthewmueller, line 11: Thanks for adding this!

.envrc
wheinze: yep that's possible via `has` directive. added a check ;) https://github.com/livebud/bud/pull/393/commits/0bcfd6ccdc598ae3940196406135289653fbcde7

note: the related docs can be found over here: https://direnv.net/man/direnv-stdlib.1.html

wheinze:
What also might be interesting in advance is, that we could easily integrate direnv with `.env` (dotenv) files by using [`dotenv_if_exists`](https://direnv.net/man/direnv-stdlib.1.html#codedotenvifexists-ltdotenvpathgtcode)

Recently, I also stumbled upon https://github.com/jdxcode/rtx which look quite promising to me, but afaik it handles env-specific vars differently and preferably would not be used alongside with direnv. So this might be sth to consider for improvements in the future ;)

.envrc
matthewmueller: Neat!

matthewmueller:
> What also might be interesting in advance is, that we could easily integrate direnv with .env (dotenv) files by using [dotenv_if_exists](https://direnv.net/man/direnv-stdlib.1.html#codedotenvifexists-ltdotenvpathgtcode)

I think for an app this might be good, but for an open source project, there's no actually many environment variables, though I'm prototyping environment support this week for Bud 😄

---

Let's give this a shot either way. Thanks @wheinze!

matthewmueller:
Hmm... just tried it. Getting:

```
direnv: loading ~/dev/src/github.com/livebud/bud/.envrc                                                        
direnv: using asdf
/bin/bash:1033: use_asdf: command not found
```

I'd like this to support the following scenarios, basically:
- no direnv, no problem
- no asdf, no problem
- direnv, no asdf, no problem
- asdf, no direnv, no problem

Where asdf & direnv are progressive enhancements. Reverting for now, sorry.

